PR #8, opened 2018-09-27
convert links to xrefs (1.x)

mojavelinux:


mojavelinux:
This one looks a bit bizarre because api-manager got split from a versionless into a versioned component. So some pages are in one version and some in the other. Until it settles, it's best to just refer to the version explicitly.

sleslie23 (approved):

PR #15, opened 2018-10-25
Update develop-custom-policies-reference.adoc

gustavo-caracoglia:
DOCS-3203 - Based on conversation with API Manager team, custom policies YAML tags "requiredCharacteristics" and "providedCharacteristics" are deprecated for Mule 4 and should not be used. After pressing API Manager "Add Policy" button, an error is thrown as well due to this.


sleslie23:
@gustavo-caracoglia Can you resubmit this PR into the correct versioned branch?

PR #19, opened 2018-10-26
Update oauth2-policies-new.adoc

mmartinez31:
In the line "Like other API Manager-enforced policies, the API needs to be registered in API Manager to apply and OAuth 2.0 Access Token policy.," I believe the "and" has to be replaced with "an."


chris-campbell9:
This page has been deprecated and replaced with newer content.

PR #25, opened 2018-11-15
Policy id is missing

fribasmule:
Policy ID is missing in the YAML template. If the customer doesnt add this setting in yaml , they will need to harcode it on policy xml, but in the documentation is calling a property

<policy xmlns="http://www.mulesoft.org/schema/mule/policy"
        id="{{policyId}}"
        policyName="Regex Filter"


glenn-rodgers-sf:

PR #32, opened 2018-12-04
add note about response timeout for report API

alexsmirnov:
Notes about response timeout and possible data truncate.

### DIFF
--- a/modules/ROOT/pages/analytics-event-api.adoc
+++ b/modules/ROOT/pages/analytics-event-api.adoc
@@ -17,6 +17,7 @@ The Analytics Event API enforces a usage limit of 10 requests/minute, with a max
 ** Retrieve all the data that you need in a single request. Do not repeat the same request for every API that you have.
 ** Do not call Event API more than once per minute.
 ** For organizations whose API usage exceed 20 000 calls per 10 minutes, consider a different way for data export. API gateway can be configured to send usage events directly to monitoring systems.
+** Response timeout set to 5 minutes. Requests that take more than 5 minutes to process data will be truncated. The most common cases for long requests are api / api version filters.
 * Event API is not meant for data synchronization. +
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/analytics-event-api.adoc
```diff
@@ -17,6 +17,7 @@ The Analytics Event API enforces a usage limit of 10 requests/minute, with a max
 ** Retrieve all the data that you need in a single request. Do not repeat the same request for every API that you have.
 ** Do not call Event API more than once per minute.
 ** For organizations whose API usage exceed 20 000 calls per 10 minutes, consider a different way for data export. API gateway can be configured to send usage events directly to monitoring systems.
+** Response timeout set to 5 minutes. Requests that take more than 5 minutes to process data will be truncated. The most common cases for long requests are api / api version filters. Reduce the query time rage.
```
fermujica, line 20: The message is not very clear. I would reword to something like this:

> Event API has a five minutes response timeout. If you download a report through a request that takes more than 5 minutes to process, your results will be truncated because of the timeout.  +
> API and API Version filters are a common cause for long request processing. 

Add the `+` character so the common causes are displayed after a line break.  

I'm not sure what you mean by the `Reduce the query time rage` sentence. Is this something that the user can edit from the filters?

fermujica (changes requested):

modules/ROOT/pages/analytics-event-api.adoc
pengjasonmule: Hello @fermujica ,Yes the reduce query time ra[n]ge can be edited by user (missing an n there). I would rephrase the sentence to "If you observe that data has been truncated, please reduce the search time range"

glenn-rodgers-sf:

PR #38, opened 2018-12-20
Docs-feedback fixes

george-hoffer:
See https://mulesoft.slack.com/archives/C02HGAD2E/p1545335430082500

### DIFF
--- a/modules/ROOT/pages/mule-oauth-provider-landing-page.adoc
+++ b/modules/ROOT/pages/mule-oauth-provider-landing-page.adoc
@@ -16,6 +16,11 @@ Mule OAuth 2.0 Provider is an OAuth 2.0 provider alternative developed by MuleSo
 
 In addition to Mule OAuth 2.0 Provider, Mulesoft offers out-of-the-box integration with PingFederate, OpenAM and Open ID Connect OAuth 2.0 Providers.
 
+*Notes:*
+
+*  Mule OAuth 2.0 Provider is not an actual OAuth provider and is not meant to take the place of an actual identity provider.
+* The OAuth Provider API does have a supported service level agreement.
+
 == See Also
 *** xref:oauth2-provider-configuration.adoc[Mule OAuth 2.0 Provider Configuration]
 *** xref:about-configure-api-for-oauth.adoc[OAuth 2.0 Policy Prerequisites]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/mule-oauth-provider-landing-page.adoc
```diff
@@ -18,7 +18,7 @@ In addition to Mule OAuth 2.0 Provider, Mulesoft offers out-of-the-box integrati
 
 *Notes:*
 
-*  Mule OAuth 2.0 Provider is not an actual OAuth provider and is not meant to take the place of an actual identity provider.
+*  Mule OAuth 2.0 Provider is not an actual OAuth provider and is not meant to take the place of an actual identity provider. You can still use OAuth 2.0 Provider in production but not as your *only* OAuth Provider. You still need some sort of store or IDP in the backend while the provider can be how your application integrates with the Mule platform.
```
Mulesoft-Mysti: Revise for clarity, concision? Such as:

You can use OAuth 2.0 Provider in production to integrate directly with Anypoint Platform. However, you also need an IdP or store to manage (back-end transactions? Authorization into your production environment?). Why would anyone do that? Should we just go with the original comments?

Mulesoft-Mysti (changes requested):

modules/ROOT/pages/mule-oauth-provider-landing-page.adoc
george-hoffer: I don't think Mule platform and Anypoint Platform are the same. Mule platform is more of a catch-all for Mule on prem and cloud, and standalone or via Anypoint Platform.
How about:
OAuth 2.0 Provider should not be the only OAuth provider. Use this product in conjunction with an identity provider to handle authorization in a Mule application environment.

Mulesoft-Mysti: There is no such product as Mule platform so we can't say that. You can say Mule to mean all those things. Your rewrite feels less specific. Do apps integrate with Mule? I thought they used Mule to integrate with each other.  So are we saying you can't use it between apps, but can to manage identity between Mule and an app that you're integrating with Mule (but not app-to-app?).

evangelina: @chiragnarangmulesoft , could you please review the statement around OAuth 2.0 provider in production and confirm?

glenn-rodgers-sf:

PR #45, opened 2019-01-03
Fixed the id of the maven repository

adampridmore:
This matches the id's created in the archetype for the custom policy.


chris-campbell9:
Fixed in the following PR:
https://github.com/mulesoft/docs-api-manager/pull/181

PR #59, opened 2019-02-05
suggest clarifying "Identifier" in RL policy

PaulAnderson-One:
@fermujica tagging you as requested

See https://mulesoft.slack.com/archives/C02HAAH50/p1549394617203100 for context.

  - the word "Identifier" doesn't appear anywhere in this doc-page. The utility of the "Identifier" and how to actually use it isn't identified at all; it's understood that the identifier identifies a bucket; what's missing is a sample usage to bridge the gap between the notion of a buckets subdividing the allocation, and what was envisioned as passing as the identifier qparam.

- also missing is what happens if (DoS scenario) someone makes 1 000 000 requests - within the RL bounds - with 1 000 000 different identifier values. Maybe it's appropriate to advise users "include identifier qparam in the RAML, and constrain it with an enum")
  - the fact that "Identifier" is optional probably also needs to be explained. Seems that with no "identifier" specified, all requests count against the same (ie default) "bucket".


glenn-rodgers-sf:

PR #64, opened 2019-02-18
Update available-policies.adoc

PabloCamino:
Added clarification on applying same policy more than once according to SE-10851


glenn-rodgers-sf:

PR #67, opened 2019-02-19
Add list of supported keystores.

fermujica:
Adding a list of supported keystores. 

TLS contexts using PEM types keystores cannot be selected when the secret groups' target is of type Mule.


sbehrends (approved):
Confirmed with @adcarabajal.

renukapatil17:
Confirmed.

PR #76, opened 2019-03-20
Add more information about x-agw headers in federation policies

fermujica:


modules/ROOT/pages/oauth-policy-implementation-concept.adoc
```diff
@@ -66,17 +66,26 @@ image::bad-request.png[]
...
+#[authentication]
+--
 
-`&#x0023;[authentication]`
+To retrieve the UID passed in `x-agw-uid`, use:
 
-To retrieve the UID, use:
+[source,dataweave,linenumes]
+--
+#[authentication.properties.userProperties.username]
```
fermujica: Is this expression correct to retrieve the UID? 
It does not seem consistent with what it's being said below:
> `X-AGW-{key}=value` header, using the expression `#[authentication.properties.userProperties.{key}]`

balbifm: No, for username is `authentication.properties.userProperties.username` and for UID is `authentication.properties.userProperties.uid`, so both expressions should be updated

fermujica: Thanks! It makes much more sense now.
6e9af28

balbifm (approved):

Cristian-Venticinque (approved):

PR #80, opened 2019-03-29
API visibility restricted to env in crowd2

rodrigo-neira:
Many cases were raised because customers want to restrict visibility to APIs per user, and not environment. But this was changed after new crowd release. We need a disclaimer so we reduce the ammount of cases being raised by our customers.


fermujica (approved):

fermujica:
Thanks! I added ecadaa0 to apply style to your NOTES,  and 4dd4ba9 to fix a broken anchor link I spotted in this article.

PR #98, opened 2019-05-07
Add information on how to download the Mule OAuth provider

fermujica:


mtqp (approved):

modules/ROOT/pages/oauth2-provider-configuration.adoc
```diff
@@ -3,39 +3,51 @@ ifndef::env-site,env-github[]
...
 You can define scopes in three different places.
 
-First, the universal set (ie "defaultScopes"), where you define all possible scopes.
+First, the universal set (ie `defaultScopes`), where you define all possible scopes.
 
-Then, you may configure the validation endpoint (eg /validate) to enforce a particular subset of the the scopes defined in the previous step.
+Then, you may configure the validation endpoint (eg `/validate`) to enforce a particular subset of the previously defined scopes.
 
-You may also enforce scopes by specifying a space separated list in the policy itself.
+You may also enforce scopes by specifying a space-separated list in the policy itself.
```
Cristian-Venticinque, line 59: Is there any difference between these three places where the scopes can be define? Do we suggest to use the first, or each option is used with a different objective?

Otherwise, these could be just bullet points without specifying order, for example: 

You can define scopes in three different places:
* In the universal set (ie `defaultScopes`), where you define all possible scopes.
* In the validation endpoint (eg `/validate`), to enforce a particular subset of the previously defined scopes.
* In the policy, by specifying a space-separated list.

fermujica: Good point. Your suggestion makes more sense:
089dd69

PR #99, opened 2019-05-07
DOCS4874 Adding suffixes to titles about APID in API Manager

rddheath:
The point of adding suffixes to the titles is to distinguish these files from those for APID in Design Center.

### DIFF
--- a/modules/ROOT/pages/api-manager-designer-archive.adoc
+++ b/modules/ROOT/pages/api-manager-designer-archive.adoc
@@ -1,4 +1,4 @@
-= API Designer
+= API Designer (in API Manager)
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
--- a/modules/ROOT/pages/api-manager-designer-archive.adoc
+++ b/modules/ROOT/pages/api-manager-designer-archive.adoc
@@ -1,4 +1,4 @@
-= API Designer (in API Manager)
+= API Designer (API Manager)
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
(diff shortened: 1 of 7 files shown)


fermujica:
I assume these are useful when the titles show up as search results. 
QQ: Is Design Center (mostly flow designer) available in the pre-Crowd release of the platform? Because API Manager v1.x refers to that previous version of the platform). I'd like to know if it's really necessary to make that distinction. 
In the context of the tile of the article, it might be redundant, since the API Manager menu in the TOC should be selected when the users read this article.

rddheath:
@fermujica , yes, I'm thinking of search results. My concern is that, say, users of APID in Design Center will use our Coveo search or a Google search to find the documentation for APID in DC, but will instead see results for APID in API Manager. The same goes the other way round: users of APID in APIM seeing results for APID in Design Center. This problem could happen whether or not DC is available in the pre-Crowd release. I'm not sure why the question of pre-Crowd versus Crowd is relevant for products that can equally be found in searches.
Also, I don't think that Design Center is "mostly" Flow Designer. DC contains both APID and Flow Designer, and I wouldn't say that FD is bigger than APID or that it is presented that way. 
About your last point: the title of the article is what appears in search results. So, whether or not "API Manager" is visible in the TOC when a topic is opened is beside the point. Moreover, when I've searched on APID and opened up topics that are in the API Manager docs, the 1.x docs for APIM are not shown in the TOC; the current topic is not selected in the TOC until after I select the 1.x version number next to the API Manager entry in the TOC.

fermujica:
I see your point. 
We do the same to differentiate between deployment tasks from Studio vs other deployment methods such as the Mule Maven Plugin, or the Runtime Manager UI.

```
= Deploying a Mule Application to CloudHub (Anypoint Studio)
```
vs. 
```
= Mule Maven Plugin
...
== Deploy a Mule Application to CloudHub
```

To keep consistency with this method, could you remove the `in` and `for`s in the titles? 
```
= API Designer (API Manager)
```

fermujica (approved):
Thanks for making the change!

PR #100, opened 2019-05-09
[DO NOT MERGE] AGW-2820: Document the oauth2 provider in gateway mode for mule 4.x

Pgarrett:


mtqp:
Keywords and sentences need to be included:
- This is the Client Store that API Gateway provides (requires API Gateway license and startup credentials --> add links to that)
- It is used to create tokens for Client Applications defined in API Platform.
- Every request is validated against the platform, though it uses a local cache for when there's no connectivity. The cache can be persistent (explain the property).

PR #101, opened 2019-05-09
(do not merge yet) remove AWS URLs from whitelist

wwarner-salesforce:
Per https://www.mulesoft.org/jira/browse/EXC-2680 and https://mulesoft-roadmap.aha.io/epics/EX-E-170 , API policies now come from a MuleSoft domain URL and not an Amazon S3 domain URL. Before merging, confirm with Kuntal and possibly other Exchange team members that this is correct, and confirm with Fernando or API Manager team that the Amazon domains are not needed on the whitelist for any other purpose.

### DIFF
--- a/modules/ROOT/pages/runtime-urls-whitelist.adoc
+++ b/modules/ROOT/pages/runtime-urls-whitelist.adoc
@@ -17,7 +17,6 @@ The following table shows you the ports or IPs/hostnames to add to your whitelis
 |*US*|*arm-auth-proxy.prod.cloudhub.io* |  443
 |*US*|*data-authenticator.anypoint.mulesoft.com* |  443
 |*US*|*exchange-files.anypoint.mulesoft.com* |  443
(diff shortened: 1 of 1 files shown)


kuntald:
@wwarner-salesforce , @mcampo :
I don't think we should "remove" the AWS URLs from the whitelist, but we should add the new one, if there is one, that has the mulesoft vanity domain URL. The reason being, existing portal pages for customers may still contain images with the old AWS URLs, and we need these URLs to be still whitelisted, for backward compatibility.

What we SHOULD however, also document, is the steps using which customers can update the old Portal Images to use the new URL, if they so choose to.

wwarner-salesforce:
Ah, good point. I'll close this PR.

My understanding is that https://www.mulesoft.org/jira/browse/EXC-3394 for enabling secure portal images is not released yet but will be soon, so I'll work on drafting that documentation.

PR #105, opened 2019-05-20
Update ip-whitelist.adoc

angelspano:
Please accept the change is just a wrong single quote
#[message.inboundProperties['X-Forwarded-For’]]

for 
#[message.inboundProperties['X-Forwarded-For']]


angelspano:
Having second thoughts on the Mule 4 expression, if you check the IP Blacklist, the expression is different:
#[message.inboundProperties['http.headers']['X-Forwarded-For']]

chris-campbell9:
Fixed in the following PR:
https://github.com/mulesoft/docs-api-manager/pull/123

PR #110, opened 2019-05-30
 Adding platform credentials encrpyiton for Gateway mule 3

fermujica:


modules/ROOT/pages/api-gateway-encryption-mule3.adoc
```diff
@@ -0,0 +1,113 @@
...
+O1q1lBrvkfOVK0druUcFWzLmcmXHWr3tFKuFqucze8mZ7QljItYaxF+iZlgIRzhCkd2q95u9zOif
+wVmIJxk1kv4Dha5jQM15uKP/22fMrMkKljGEhApKlhBwevlb542TeXWETlaRU8sZpXLscrLPCHiU
+XskWI+xK19uVgofmK88iyhWr3gdYY5Aefvdj6Lxr/MMVgve4kYpd7jJpYycS9GzObPIyaOEvL5DG
+Hvw=]
+----
+
+If you need to debug or troubleshoot your policies configuration, you can use the policy mode of the gateway encryption tool to recover the non-encrypted XML.
+
+The runtime does not automatically encrypt offline policies. However, if you provide the offline policy files already encrypted, the runtime is able to decrypt them and use them normally. +
+When running in clusters, offline policies remain encrypted in the node in which the policy was originally deployed, but it will propagate unencrypted to the other nodes.
```
evangelina: Shouldn't this mention that if the other nodes are configured with encryption it will be stored encrypted? @JaatadiaMulesoft

JaatadiaMulesoft: Yes we should. "offline policies remain encrypted in the node (If that node was configuread with encryption)..."

evangelina: Thanks for confirming. @fermujica could we make this small change? Other than that it looks good to merge.

fermujica: Would this make more sense?
31ae0a6

modules/ROOT/pages/api-gateway-encryption-mule3.adoc
```diff
@@ -0,0 +1,114 @@
...
+[source]
+----
+$MULE_HOME/bin/mule start -Danypoint.platform.encryption_key=MyEncryptionKey1
+----
+
+While the runtime is starting it checks for an encryption key. Then it gets the client ID and client secret values from its `wrapper.conf` file.
+
+== Encrypting Credentials using the Agent
+
+To configure the runtime through the agent to use encryption the argument `encryption-key` with the desired key as value. The agent then encrypts the client ID and Secret.
```
Cristian-Venticinque: There's something missing in this sentence: "To configure the runtime through the agent to use encryption the argument `encryption-key` with the desired key as value." 
Please check if this is the intended meaning: 
"To configure the runtime through the agent to use encryption, the argument `encryption-key` **must be configured** with the desired key as value.

Moreover, you could separate the runtime and the agent configuration to clarify: 
"To configure the runtime to use encryption, set the argument `encryption-key` with the desired key as value in the agent." or something like that.

modules/ROOT/pages/api-gateway-encryption-mule3.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+When starting the runtime, you need to pass the encryption key using the `anypoint.platform.encryption_key` property as instructed above.
+
+== Gateway Encryption Tool
+
+The gateway encryption tool provides a way to inspect the policy files that Mule runtime 3.9.3 and later encrypts, provides a way to encrypt offline policies, and pass encrypted client credentials to the runtime.
```
Cristian-Venticinque: "...and later encrypts. It also provides a way..."

modules/ROOT/pages/api-gateway-encryption-mule3.adoc
```diff
@@ -0,0 +1,114 @@
...
+java -jar gateway-encryption-tool.jar property <encrypt|decrypt> <key> <input_value>
+----
+
+This tool takes the key passed in `<key>` and use it to encrypt the value passed in `<input_value>`.
+
+When using the `decrypt` option, it is possible that your terminal might interpret some of the characters of the encrypted value. To avoid this scenario, wrap the encrypted value in simple comas:
+
+[source]
+----
+java -jar gateway-encryption-tool.jar property decrypt 1234567812345678 ‘![FefVUnvDJpbXnUvrGAEezg==]’
```
Cristian-Venticinque: is `’` the proper character here? It is referred as `simple coma`, but the name of such character seems to be `inverted single comma` which doesn't seem to be the same character as the `single quotation mark`: `'`

Can you please confirm that the character used here is correct, and also confirm its proper name? Having the unicode of that character might help identify it.

Cristian-Venticinque (approved):

PR #141, opened 2019-09-11
AGW-3394: Make enabling SSL Validation configurable in Federation policies

NachoRaik:


priyaayyar (approved):
Made changes directly to the content.

PR #160, opened 2019-11-08
DOCS-5648-applied-template-to-policy

priyaayyar:


modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -108,17 +120,22 @@ curl "http://localhost/myResource" -u 1234:abcd
 
 Example DataWeave 2.0 expression to be used when configuring the policy:
 
-No DataWeave 2.0 expression needed to be specified when the policy is configured to use the Basic Authentication as the origin of the credentials.
+To use the Basic Authentication as the origin of the credentials, no DataWeave 2.0 expression needs to be specified when configuring the policy.
 
-== WWW Authenticate HTTP header
-When a HTTP request is performed against a protected resource, where the Client ID enforcement policy is applied, and client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
+==== With WWW Authenticate HTTP header
```
JaatadiaMulesoft: This is not another way of configuring the policy, it's an effect that is present no matter how you configure the policy. So, if you apply a client Id based policy and you don't provide the credentials when making the request, the response will have the 'WWW-Authenticate' header.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -108,17 +120,22 @@ curl "http://localhost/myResource" -u 1234:abcd
...
+When an HTTP request is performed against a protected resource for which the Client ID Enforcement policy is applied, and client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
 
 [width="100%", cols="2,70"]
 |===
 >s| Custom mode | Header 'WWW-Authenticate'='Client-ID-Enforcement'
 >s| Basic Auth mode | Header 'WWW-Authenticate'='Basic realm="mule-realm"
 |===
 
+
+=== Configure API Specifications
```
JaatadiaMulesoft: Due to what i stated in the previous comment this section here shouldn't exist.

JaatadiaMulesoft (dismissed):

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
 |500 - Bad response from authorization server, or WSDL SOAP Fault error
 |===
 
-The Client ID Enforcement Policy restricts access to a protected resource, by only allowing HTTP requests if credentials of client applications already xref:api-contracts-landing-page.adoc[registered to (contracted with) the API being targeted] are provided in each request.
+The Client ID Enforcement Policy restricts access to a protected resource, by only allowing HTTP requests if credentials of client applications already xref:api-contracts-landing-page.adoc[registered to (contracted with) the API being targeted] are provided in each request. In Anypoint Platform, the client ID and client secret pair is also known as credentials.
```
evangelina, line 18: Is there a policy around having the links embedded or in the see more section? Mainly asking about the xref here.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
```
evangelina: I would rephrase this a bit:

The Client ID Enforcement Policy restricts access to a protected resource by only allowing requests from registered client applications. The policy ensures that the client credentials sent on each request have been approved to consume the API.

When a client application is registered in Anypoint Platform, a set of credentials consisting of a client id and client secret pair are granted. Once the client application requests access to an API, a contract is generated between the application and the API. An API that is protected with a Client ID Enforcement policy will only be accessible by applications that have an approved contract. 

(In the see more section we link to all the client credentials plus contracts documentation)

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
 
-The policy validates the Client ID and, optionally, the Client Secret, of a client application created within an Anypoint Platform Organization.
+After the Client ID Enforcement policy is applied, the Mule runtime engine (Mule) caches the information locally and updates it periodically. In addition to performance improvements, caching enables the policy to work offline during connectivity outages. 
 
-IMPORTANT: This policy will not validate successfully any client application credentials generated outside of the Anypoint Platform context.
+Regardless of the state of the management plane, the credentials are enforced with the existing data. Once the connectivity is re-established, the cache resumes periodical updates.
 
-NOTE: The set of Client ID and Client Secret are also known as credentials.
+//What information is stored or retrieved. -- Eva to send
+//Latest version has encryption. So the information is encrypted.
```
evangelina: I was thinking that we might not want to go into details of what is stored, mainly for security reasons. So I wouldn't expand much more on it. 

About encryption,. I have the same doubt. but if we were to put a paragraph about it, it could be a note + reference docs. 

Note: Mule Runtime can be configured to encrypt all sensitive information related to policies, contracts and initialization data.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
 
-= Analytics and  access tracking
```
evangelina, line 25: Applying a client if enforcement policy will also enable access tracking of your API by reporting the client id with the analytics events.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
-If you need to track client application access to your API, apply a policy that directly or indirectly makes sure that client ID is provided in the HTTP request (ie Client ID enforcement type policy).
+// How does the policy work?
+//Where can you get the inforamtion for the policy--Mariano to send me the information
 
-= Indirect client ID enforcement
-There are other provided policies that enforce client application credentials. For example:
 
-* xref:rate-limiting-and-throttling-sla-based-policies.adoc[Rate Limiting - SLA-Based Policies]
+Default policies that internally enforce client application credentials include:
```
evangelina, line 23: There are other default policies that internally enforce requests coming from registered client applications:

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
 
+* xref:rate-limiting-and-throttling-sla-based-policies.adoc[Rate Limiting - SLA-Based Policies]
++
 In Mule 3.8.5 or later, the following OAuth 2.0 Token Enforcement policies include a client ID enforcement validation:
-
 * xref:external-oauth-2.0-token-validation-policy.adoc[Mule OAuth 2.0 Access Token Enforcement]
 * xref:policy-ping-federate.adoc[PingFederate OAuth 2.0 Token Enforcement]
 * xref:policy-openid-connect.adoc[OpenID Connect OAuth 2.0 Token Enforcement]
```
evangelina: JWT Policy that is supported since Mule 4 is missing

evangelina: I would also add a note at the end. Even though the token enforcement policies have an option to skip client validation, the recommendation is to enforce the validation to ensure that the token is associated to an approved contract.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
 
-= Offline mode
+== Policy Prerequisites and Restrictions
 
-Credentials data from client application credentials registered to protected APIs, are provided by Anypoint Platform. Once a policy requiring such information is applied (eg Client ID enforcement policy), the Runtime caches persistently the information locally updating it periodically.
+For the policy to work successfully, you must:
```
evangelina: I don't think there is any prerequisite for the policy or restrictions. But if you want to have a section of "how the policy works":

Every client application that wants to consume an API protected by a Client ID enforcement policy needs to previously have requested access to this API. Once there is an approved contract between the application and the API, every request needs to include the client credentials in compliance with how the policy has been configured. 

As an example, if the policy has been configured to expect a client id and client secret as headers, the application needs to make sure the credentials are sent in the request using the corresponding headers. 

I would also have a different section for "API Specification" that includes the rest console information:

The client id enforcement policy will require changes in your API specification to reflect the need to provide the credentials. 
You can find the RAML/OAS snippet link containing the RAML/OAS code you need to add to the API spec in the corresponding policy from the list of applied policies in the Policies tab of your API specification in API Manager.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
+To obtain the RAML code to add to the policy, go to the corresponding policy in your API Manager.
+//It looks like we might need this for other policies too? If so, we should move it out of here and put it in a central place.
++
+image::raml-snippet.png[]
 
-= REST Console
-To prevent user requests from being rejected when using the REST Console, create a trait in the RAML root and then reference it in every operation of your API affected by policy. You may find the RAML snippet link containing the RAML code you need to add to the RAML in the corresponding policy from the list of applied policies in the Policies tab of your API specification in API Manager.
 
-image::raml-snippet.png[]
+== Policy Implementation Best Practices
```
evangelina: Agree, I wouldn't have a section just for this now.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
+
+//This may be too little info to warrant a section of its own. Let's reassess.
 
-= Credentials source
 
-The policy can be configured to extract the client ID and client secret from the HTTP request in different ways using DataWeave 2.0 expressions.
+== Configure the Policy
 
-The following examples assume that the values for application credentials are '1234' for client ID and 'abcd' for client secret.
+You can configure the policy to extract the client ID and client secret from the HTTP request in different ways using DataWeave 2.0 expressions. If when configuring the policy, the client secret DataWeave 2.0 expression is not specified, then only the client ID specified in the HTTP request is validated against the protected resource.
```
evangelina: I would rephrase this one, the section will be Configure the policy and then the different ways to do it right? Then I suggest:

The Client ID enforcement policy can be configured to obtain the client credentials in different ways. One option is to configure the policy to expect the credentials as part of the Authorization header by selecting the HTTP Basic Authentication option, in this case the application consuming the API needs to use basic authentication to send the credentials in the requests. 

The other alternative is to select Custom Expression, this is the default selection and it allows the customer to provide an expression to decide where to extract the credentials from the request. In Mule 3 a customer can provide a MEL Expression, in Mule 4 it should be a DataWeave 2.0 expression [not sure if this page is just for 4, 3 or shared). This option is needed when the customer wants to send the credentials in a custom header or somewhere else in the payload. 

There is also the option to set the client secret as optional by leaving the field blank. 

Following are some configuration examples:

[I would leave the invalid scenario after the configuration examples]

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
 
-NOTE: If when configuring the policy, client secret DataWeave 2.0 expression is specifically not specified (ie empty field), then only the client ID specified on the HTTP requests against the protected resource will be validated.
+When an HTTP request is performed against a protected resource for which the Client ID Enforcement policy is applied, and the client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
```
evangelina: This should probably go in a Policy Response section. 

Note: With the exception of SOAP applications where the response will be a SOAP Fault.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,73 @@ endif::[]
...
+
+[width="100%", cols="2,70"]
+|===
+>s| Custom mode | Header 'WWW-Authenticate'='Client-ID-Enforcement'
+>s| Basic Auth mode | Header 'WWW-Authenticate'='Basic realm="mule-realm"
+|===
 
-== Credentials as HTTP headers
+
+=== Configure Parameters
```
evangelina: I would remove this and have the sections as part of the examples of the configuration of the policy

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -70,7 +95,7 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
 #[attributes.headers.['client_secret']]
 ----
```
evangelina, line 63: In this example the policy is configured to expect two headers: client_id and client_secret with the pair of credentials. The policy is flexible to allow other type of headers as well. 

This is the default configuration for the policy.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -85,7 +110,7 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
 #[attributes.queryParams.'client_secret']
 ----
```
evangelina, line 81: Note that in this case the requester needs to send the two specified query parameters. Although this is a supported configuration, it is not recommended to send the credentials as query parameters since this could pose a security risk. The recommended practice is to use headers.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -85,7 +110,7 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
 
-== Credentials included in HTTP request payload
+==== With HTTP Request Payload
 Example request using curl:
 [source,text]
 ----
```
evangelina, line 88: It could also be possible to configure the policy to obtain the credentials from the request payload like in this example. Although feasible, it is not a recommended practice since it is harder to reflect in the API specification.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -108,23 +133,19 @@ curl "http://localhost/myResource" -u 1234:abcd
...
-== WWW Authenticate HTTP header
-When a HTTP request is performed against a protected resource, where the Client ID enforcement policy is applied, and client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
 
-[width="100%", cols="2,70"]
-|===
->s| Custom mode | Header 'WWW-Authenticate'='Client-ID-Enforcement'
->s| Basic Auth mode | Header 'WWW-Authenticate'='Basic realm="mule-realm"
-|===
+=== Configure API Specifications
+//What to add here? -- Eva (this is not needed?)
```
evangelina: It's already covered earlier in the doc, I would leave it for now and marked it as one of the improvements areas in the documentation: how to update API spec once a policy is applied.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -108,23 +133,19 @@ curl "http://localhost/myResource" -u 1234:abcd
...
 == See Also
 
 * xref:rate-limiting-and-throttling-sla-based-policies.adoc[Rate Limiting - SLA-Based Policies concepts]
 * xref:policy-ping-federate.adoc[PingFederate OAuth 2.0 Token Enforcement]
 * xref:policy-openid-connect.adoc[OpenID Connect OAuth 2.0 Token Enforcement]
-* xref:external-oauth-2.0-token-validation-policy.adoc[Mule OAuth 2.0 Access Token]
+//* xref:external-oauth-2.0-token-validation-policy.adoc[Mule OAuth 2.0 Access Token]
 * xref:defining-sla-tiers.adoc[Reviewing SLA Tiers concepts on API Manager]
 * xref:manage-client-apps-latest-task.adoc[Managing API instance Contracts on API Manager]
 * xref:access-client-app-id-task.adoc[Getting Client ID and Client Secret of a registered Client App]
```
evangelina: I would order them differently, all the contract information first, since that is crucial for the workflow of this policy, and then you can add the links to the other policies that do internal enforcement of clients as well.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Moved it to See Also.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Great! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: I added the note. What link to add?

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Which? We should name them all.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Added a link! Thanks!

priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-The policy validates the Client ID and, optionally, the Client Secret, of a client application created within an Anypoint Platform Organization.
+When a client application is registered in Anypoint Platform, a set of credentials consisting of a client id and client secret pair are generated. Once the client application requests access to an API, a contract is created between the application and that API. An API that is protected with a Client ID Enforcement policy will only be accessible to applications that have an approved contract.
 
-IMPORTANT: This policy will not validate successfully any client application credentials generated outside of the Anypoint Platform context.
+When you apply a Client ID Enforcement policy, access tracking of your API is enabled by reporting the client ID along with the analytics events. You can configure Mule runtime engine (Mule) to encrypt all sensitive information thar pertains to policies, contracts, and initialization data.
```
evangelina: typo in thar. There is no link on "See more" for encryption.

modules/ROOT/pages/client-id-based-policies.adoc
evangelina: JWT was the one to add, I added in the comment and I can see you added it in the list.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 image::raml-snippet.png[]
 
-= Credentials source
+== Configure the Policy
+
+You can configure the Client ID Enforcement policy to obtain the client credentials in two different ways, HTTP basic authentication header and custom expression. 
+
+*HTTP Basic Authentication Header*
+
+You can configure the policy to require credentials as part of the Authorization header by selecting the HTTP Basic Authentication option. In this case, the application consuming the API must use the basic authentication to send the credentials in the requests.
```
evangelina: replace "the basic authentication" with "the basic authentication scheme"

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-The policy can be configured to extract the client ID and client secret from the HTTP request in different ways using DataWeave 2.0 expressions.
+*Custom Expression*
 
-The following examples assume that the values for application credentials are '1234' for client ID and 'abcd' for client secret.
+This options allows you to provide an expression each for client ID and client secret, specifying where to extract the credentials from the request. In Mule 3 you can provide an MEL Expression and in Mule 4 you must provide  a DataWeave 2.0 expression. Use this option when you want to send the credentials in a custom header or somewhere else in the payload. You can also configure the client secret requirement as optional by leaving the field blank.
```
evangelina: I'm not an english native, but is "allows you to provide an expression each for client ID and client secret" correct?

evangelina: There is an extra space in "must provide a DataWeave 2.0 expression"

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -108,23 +121,32 @@ curl "http://localhost/myResource" -u 1234:abcd
 
 Example DataWeave 2.0 expression to be used when configuring the policy:
 
-No DataWeave 2.0 expression needed to be specified when the policy is configured to use the Basic Authentication as the origin of the credentials.
+To use the Basic Authentication as the origin of the credentials, no DataWeave 2.0 expression needs to be specified when configuring the policy.
```
evangelina: This is a bit confusing, no expression needs to be configured because they have to select the other option (not custom headers).

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -108,23 +121,32 @@ curl "http://localhost/myResource" -u 1234:abcd
...
 |===
 >s| Custom mode | Header 'WWW-Authenticate'='Client-ID-Enforcement'
 >s| Basic Auth mode | Header 'WWW-Authenticate'='Basic realm="mule-realm"
 |===
 
+For SOAP applications, the response will be a SOAP error.
+
 == See Also
 
+* xref:manage-client-apps-latest-task.adoc[Managing API instance Contracts on API Manager]
```
evangelina, line 154: Encryption link is missing. Might be also good to add a link to the release notes on each policy, what do you think?

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Awesome! Done!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Fixed typo! Texted you about the link!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Yes for the first sentence. It means you need to provide one expression for client ID and one for client secret. Fixed the space issue. Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: This was there from before. I thought this is a requirement. If it does not make sense, should I remove it?

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Good idea! I'll add for release notes too.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
 |500 - Bad response from authorization server, or WSDL SOAP Fault error
 |===
 
-The Client ID Enforcement Policy restricts access to a protected resource, by only allowing HTTP requests if credentials of client applications already xref:api-contracts-landing-page.adoc[registered to (contracted with) the API being targeted] are provided in each request.
+The Client ID Enforcement Policy restricts access to a protected resource by only allowing requests from registered client applications. The policy ensures that the client credentials sent on each request have been approved to consume the API.
```
kbostrom:
```suggestion
The Client ID Enforcement policy restricts access to a protected resource by allowing requests only from registered client applications. The policy ensures that the client credentials sent on each request have been approved to consume the API.
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-The policy validates the Client ID and, optionally, the Client Secret, of a client application created within an Anypoint Platform Organization.
+When a client application is registered in Anypoint Platform, a set of credentials consisting of a client id and client secret pair are generated. Once the client application requests access to an API, a contract is created between the application and that API. An API that is protected with a Client ID Enforcement policy will only be accessible to applications that have an approved contract.
```
kbostrom:
```suggestion
When a client application is registered in Anypoint Platform, a pair of credentials consisting of a client ID and client secret is generated. When the client application requests access to an API, a contract is created between the application and that API. An API that is protected with a Client ID Enforcement policy is accessible only to applications that have an approved contract.
```

kbostrom: - Reworded to not have both "a set" and a "pair" in the same sentence.
- Because "once" has more than one meaning, I typically suggest "after" or "when," depending on the intent.
- Take a look at the style reference entry for "only."

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-IMPORTANT: This policy will not validate successfully any client application credentials generated outside of the Anypoint Platform context.
+When you apply a Client ID Enforcement policy, access tracking of your API is enabled by reporting the client ID along with the analytics events. You can configure Mule runtime engine (Mule) to encrypt all sensitive information that pertains to policies, contracts, and initialization data.
```
kbostrom: Should "access tracking of your API is enabled" be "access to your API is tracked"? 
```suggestion
When you apply a Client ID Enforcement policy, access to your API is tracked by client ID and analytics events. You can configure Mule runtime engine (Mule) to encrypt all sensitive information that pertains to policies, contracts, and initialization data.
```
Does your audience know what "analytics events" are? 
How does configuring Mule to protect sensitive info relate to the policy? Or is this a separate point that should maybe be in a separate para?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
 |500 - Bad response from authorization server, or WSDL SOAP Fault error
 |===
```
kbostrom, line 15: Does the title of this PR mean that you applied an existing template to a policy reference topic?
If so, should I edit that template?
I won't go into editing texts that you have not changed, since you do not have time to look into legacy tech errors, but I will make these recommendations:

- Change the capitalization of the page to use title caps: Client ID Enforcement
   And consider adding Policies to that title, so that someone skimming titles would know that this is 
   about types of policies rather than client IDs or enforcing their use.

- Is the table part of the template? Does each policy description page include it as the very first element?
   That is typically OK for a Reference topic, but this has a lot of conceptual info in it as well.

- "Since Mule Version" is bad grammar. "First Mule support" would be better, with an entry of "v3.8.0."

- Left-column entries should use sentence-style caps., because they are not titles.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
+When you apply a Client ID Enforcement policy, access tracking of your API is enabled by reporting the client ID along with the analytics events. You can configure Mule runtime engine (Mule) to encrypt all sensitive information that pertains to policies, contracts, and initialization data.
 
-NOTE: The set of Client ID and Client Secret are also known as credentials.
-
-= Analytics and  access tracking
-If you need to track client application access to your API, apply a policy that directly or indirectly makes sure that client ID is provided in the HTTP request (ie Client ID enforcement type policy).
-
-= Indirect client ID enforcement
-There are other provided policies that enforce client application credentials. For example:
+Default policies that internally enforce client application credentials include:
```
kbostrom, line 23: This sentence looks like it introduces a list, but there is just one item?
Look at how this transforms. It looks line a single-item list, with an item that has text below it and then subentries.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
 * xref:rate-limiting-and-throttling-sla-based-policies.adoc[Rate Limiting - SLA-Based Policies]
-
++
 In Mule 3.8.5 or later, the following OAuth 2.0 Token Enforcement policies include a client ID enforcement validation:
-
 * xref:external-oauth-2.0-token-validation-policy.adoc[Mule OAuth 2.0 Access Token Enforcement]
 * xref:policy-ping-federate.adoc[PingFederate OAuth 2.0 Token Enforcement]
 * xref:policy-openid-connect.adoc[OpenID Connect OAuth 2.0 Token Enforcement]
+* xref:policy-mule4-jwt-validation.adoc[JWT Validation]
```
kbostrom: Is this one also a token enforcement policy? Because it is grouped with the previous three, and also followed by text about token enforcement.
Maybe it should follow the Rate Limiting policy name?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-= Offline mode
+Although token enforcement policies can optionally skip client validation, MuleSoft recommends that you enforce the validation to ensure that the token is associated to an approved contract.
```
kbostrom: GLOBALLY reword to avoid "MuleSoft recommends" (and never use "we").

kbostrom: So, what can a user do to an existing token enforcement policy to ensure that it does NOT skip client validation?
Do they have to alter the policy, or change a policy parameter? Or do they have to take an action independent of and in addition to the token-enforcement policy?
In short, HOW to they "enforce the validation"?

modules/ROOT/pages/client-id-based-policies.adoc
kbostrom: ORGANIZATION CONSIDERATIONS
Is this a balanced and complete topic outline?

= **Client ID Enforcement**
          (table, what it does, list of/links to policies that enforce credentials)
          Does this text prepare the audience for the main subtopics of how it works and the configuration info?
  == How the Policy Works
          Is this information really different from what is currently in the topic set intro? Maybe some of that
          should be moved here?
  == Configure the Policy
          This implies a Task topic, but there are no numbered steps.
          Maybe this should really be Choosing a Policy Configuration Method?
   === Examples
          Do we really need this container level, or can we instead have four Level-3 Example topics?
    ==== Example 1
    ====2
    ====3
    ====4
   === Configure API Specifications
        Is this topic subordinate to or a peer of policy configuration?
   === Policy Output
        Ditto. If these should both be Level-2, then they too should be introduced in the first topic.
  == See Also

modules/ROOT/pages/client-id-based-policies.adoc
kbostrom: Also...this topic is about the Client ID Enforcement policy, so what are these policies and why are they listed here? Is there more than one Client ID Enforcement policy, and these are the others? 
Or are these all types of Client ID Enforcement policy?
Not sure why these are discussed as part of the topic about the (one and only?) Client ID Enforcement policy.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-Credentials data from client application credentials registered to protected APIs, are provided by Anypoint Platform. Once a policy requiring such information is applied (eg Client ID enforcement policy), the Runtime caches persistently the information locally updating it periodically.
+== How the Policy Works
 
-Besides its performance benefits, this action enables the policy to work offline in the event of a sudden loss of connectivity to the platform. The enforcement is applied regardless of the state of the management plane, with existing data. Once connectivity is re-established, the cache will be periodically updated again.
+Every client application that wants to consume an API protected by a Client ID enforcement policy needs to have previously requested access to the API. After an approved contract exists between the application and the API, every request needs to include the client credentials in compliance with how the policy has been configured.
```
kbostrom:
```suggestion
Before a client application is allowed to consume an API protected by a Client ID enforcement policy, the client must request access to the API. After an approved contract exists between the application and the API, every request must include the client credentials in compliance with how the policy is configured.
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-= REST Console
-To prevent user requests from being rejected when using the REST Console, create a trait in the RAML root and then reference it in every operation of your API affected by policy. You may find the RAML snippet link containing the RAML code you need to add to the RAML in the corresponding policy from the list of applied policies in the Policies tab of your API specification in API Manager.
+For example, if the policy has been configured to expect a client ID and client secret as headers, the application needs to make sure that those credentials are sent in the request using the corresponding headers.
```
kbostrom:
```suggestion
For example, if the policy is configured to expect a client ID and client secret as headers, the application must send those credentials in the request, using the corresponding headers:
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
 image::raml-snippet.png[]
 
-= Credentials source
+== Configure the Policy
```
kbostrom:
```suggestion
== Configuring How the Policy Obtains Credentials
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
+
+You can configure the Client ID Enforcement policy to obtain the client credentials in two different ways, HTTP basic authentication header and custom expression.
```
kbostrom:
```suggestion
You can configure the Client ID Enforcement policy to obtain the client credentials in one of two different ways: by using an HTTP basic authentication header or by using a custom expression:
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
+For example, if the policy has been configured to expect a client ID and client secret as headers, the application needs to make sure that those credentials are sent in the request using the corresponding headers.
 
 image::raml-snippet.png[]
 
-= Credentials source
+== Configure the Policy
+
+You can configure the Client ID Enforcement policy to obtain the client credentials in two different ways, HTTP basic authentication header and custom expression. 
+
+*HTTP Basic Authentication Header*
```
kbostrom:
```suggestion
* HTTP basic authentication header
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-The policy can be configured to extract the client ID and client secret from the HTTP request in different ways using DataWeave 2.0 expressions.
+You can configure the policy to require credentials as part of the Authorization header by selecting the HTTP Basic Authentication option. In this case, the application consuming the API must use the basic authentication scheme to send the credentials in the requests.
```
kbostrom:
```suggestion
You can configure the policy to require credentials as part of the authorization header by selecting the HTTP Basic Authentication option. In this case, the application consuming the API must use the basic authentication scheme to send the credentials in the requests.
```
This text uses "authorization header," "... Authentication opion," and "authentication scheme" to mean the same thing. We want to avoid confusing the audience with different names for one thing, so could we simplify this to something like "You can configure the policy to require the consuming API to provide credentials in the request header by selecting the HTTP Basic Authentication option."

But WHERE does the user select this option? Don't we need actual policy config steps that refer to the UI?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-The following examples assume that the values for application credentials are '1234' for client ID and 'abcd' for client secret.
+*Custom Expression*
 
-NOTE: If when configuring the policy, client secret DataWeave 2.0 expression is specifically not specified (ie empty field), then only the client ID specified on the HTTP requests against the protected resource will be validated.
+This options allows you to provide an expression each for client ID and client secret, specifying where to extract the credentials from the request. In Mule 3 you can provide an MEL Expression and in Mule 4 you must provide a DataWeave 2.0 expression. Use this option when you want to send the credentials in a custom header or somewhere else in the payload. You can also configure the client secret requirement as optional by leaving the field blank.
```
kbostrom:
```suggestion
This option enables you to provide an expression each for client ID and client secret, specifying where to extract the credentials from the request: either from a custom header or elsewhere in the request payload. In Mule 3 you provide an MEL Expression; in Mule 4 you provide a DataWeave 2.0 expression. 

If you prefer, you can make the client secret requirement optional by leaving its field in the configuration UI blank.
```
Again, don't we need to locate the user in the UI? Where is the user selecting options and completing fields or leaving them blank?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-== Credentials as HTTP headers
+
+=== Examples
+
+You can configure the policy to extract the client ID and client secret from the HTTP request in several ways using DataWeave 2.0 expressions. When configuring the policy, if the client secret DataWeave 2.0 expression is not specified, only then the client ID specified in the HTTP request is validated against the protected resource.
```
kbostrom:
```suggestion
You can configure the policy to extract either both the client ID and client secret or only the client ID from the HTTP request by using a variety of custom DataWeave 2.0 expressions.
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
 
-The following examples assume that the values for application credentials are '1234' for client ID and 'abcd' for client secret.
+*Custom Expression*
 
-NOTE: If when configuring the policy, client secret DataWeave 2.0 expression is specifically not specified (ie empty field), then only the client ID specified on the HTTP requests against the protected resource will be validated.
+This options allows you to provide an expression each for client ID and client secret, specifying where to extract the credentials from the request. In Mule 3 you can provide an MEL Expression and in Mule 4 you must provide a DataWeave 2.0 expression. Use this option when you want to send the credentials in a custom header or somewhere else in the payload. You can also configure the client secret requirement as optional by leaving the field blank.
 
-== Credentials as HTTP headers
+
+=== Examples
```
kbostrom:
```suggestion
== Examples of Obtaining Credentials Using Custom DataWeave Expressions
```
or
```suggestion
=== Examples of Obtaining Credentials Using Different Methods
```
The following text implies that the examples all use DataWeave custom credentials, but is that true?
Please clarify both title and text, based on the actual examples given.
Consider OMITTING ALTOGETHER the methods that we do not recommend.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
+
+You can configure the policy to extract the client ID and client secret from the HTTP request in several ways using DataWeave 2.0 expressions. When configuring the policy, if the client secret DataWeave 2.0 expression is not specified, only then the client ID specified in the HTTP request is validated against the protected resource.
+
+The following examples use client ID as '1234' and client secret as 'abcd'.
```
kbostrom:
```suggestion
The following examples use a client ID of '1234' and a client secret of 'abcd'.
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -14,48 +14,53 @@ endif::[]
...
+
+
+==== With HTTP Headers
```
kbostrom:
```suggestion
==== Obtaining Credentials Using HTTP Headers
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -70,7 +75,9 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
 #[attributes.headers.['client_secret']]
 ----
 
-== Credentials as HTTP query parameters
+In the above example, the policy is configured to expect two headers: client_id and client_secret, with the pair of credentials. The policy is flexible to allow other types of headers also. This is the default configuration for the policy.
```
kbostrom:
```suggestion
In this example, the policy is configured to expect two headers: client_id and client_secret, with the pair of credentials. The policy is flexible to allow other types of headers also. This is the default configuration for the policy.
```
How can this be the default if the policy can be configured using either basic header authentication or custom DreamWeaver expressions? Do you mean that is is the default among the custom expressions?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -70,7 +75,9 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
+
+==== With HTTP Query Parameters
```
kbostrom:
```suggestion
==== Obtaining Credentials Using HTTP Query Parameters
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -85,7 +92,10 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
 #[attributes.queryParams.'client_secret']
 ----
 
-== Credentials included in HTTP request payload
+In the above example, the requester must send the two specified query parameters with the request. Although this is a supported configuration, MuleSoft does not recommended you to send the credentials as query parameters because of possible security risks. The recommended method is to use headers.
```
kbostrom:
```suggestion
In this example, the requester must send the two specified query parameters with the request. Although this is a supported configuration, it poses possible security risks. The recommended method is to use headers.
```
I strongly recommend NOT supplying supported examples that we do not recommend.

kbostrom: I strongly recommend NOT supplying supported examples that we do not recommend.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -85,7 +92,10 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
+
+==== With HTTP Request Payload
```
kbostrom:
```suggestion
==== Obtaining Credentials Using HTTP Request Payload
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
 #[payload.client_secret]
 ----
 
-== Credentials as Basic Authorization header
+As shown in the above example, it is also possible to configure the policy to obtain the credentials from the request payload. MuleSoft does not recommend this option because it is harder to reflect in the API specification.
```
kbostrom:
```suggestion
Although you can configure the policy to obtain the credentials from the request payload, this option is not recommended because it is harder to reflect in the API specification.
```
Does your audience understand what this means: "it is harder to reflect in the API specification"?

kbostrom: Once again, I strongly recommend against providing examples of methods that we do not recommend.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
+
+==== With Basic Authorization header
```
kbostrom:
```suggestion
==== Obtaining Credentials Using the Basic Authorization Header
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
+
 Example request using curl:
 [source,text]
 ----
 curl "http://localhost/myResource" -u 1234:abcd
 ----
```
kbostrom, line 108: Where is the example of the expression used to answer the request?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
 
-Example DataWeave 2.0 expression to be used when configuring the policy:
+=== Configure API Specifications
```
kbostrom: Once again, this title implies a task topic but there are no numbered steps.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
+
+The Client ID Enforcement policy requires changes in your API specification to implement the credentials requirement.
+You can find the RAML or OAS snippet link containing the RAML or OAS code you need to add to the API specification in the corresponding policy. You can access this code from the list of applied policies in the Policies tab of your API specification in API Manager.
+//future enhancement: how to update API spec once a policy is applied.
```
kbostrom, line 138: Why include this topic at all, if HOW TO is not included?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
 
-No DataWeave 2.0 expression needed to be specified when the policy is configured to use the Basic Authentication as the origin of the credentials.
+=== Policy Output
```
kbostrom: This is misleading. Why not just call this Error Messages? But why even document these?
This text only documents the output that indicates a failure, without addressing what caused it or how to troubleshoot it. 
Of what use is this information to the audience?

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
-When a HTTP request is performed against a protected resource, where the Client ID enforcement policy is applied, and client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
+When an HTTP request is performed against a protected resource for which the Client ID Enforcement policy is applied, and the client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
 
 [width="100%", cols="2,70"]
 |===
 >s| Custom mode | Header 'WWW-Authenticate'='Client-ID-Enforcement'
 >s| Basic Auth mode | Header 'WWW-Authenticate'='Basic realm="mule-realm"
 |===
 
+For SOAP applications, the response will be a SOAP error.
```
kbostrom:
```suggestion
For SOAP applications, the response is a SOAP error.
```

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -99,32 +109,42 @@ Example DataWeave 2.0 expression to be used when configuring the policy:
...
 
-== WWW Authenticate HTTP header
-When a HTTP request is performed against a protected resource, where the Client ID enforcement policy is applied, and client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
+When an HTTP request is performed against a protected resource for which the Client ID Enforcement policy is applied, and the client application credentials are invalid or unauthorized, the HTTP response includes a WWW Authenticate header with the following values:
 
 [width="100%", cols="2,70"]
 |===
 >s| Custom mode | Header 'WWW-Authenticate'='Client-ID-Enforcement'
 >s| Basic Auth mode | Header 'WWW-Authenticate'='Basic realm="mule-realm"
 |===
```
kbostrom, line 149: "custom mode" and "basic auth mode" do not exactly match how these two are referred to earlier.
Avoid using more than one term for one thing, GLOBALLY. It confuses readers.

kbostrom (approved):
Here you go!
I made some org comments; wasn't quite sure how some of the text related to other text.
Also, I strongly recommend NOT documenting user actions that we do not recommend. Best practices only, unless we state a good reason for using such a method.
Just because users CAN does not mean that they SHOULD, and we should push back on that if nec.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Thanks for bringing these points, but we are good! Checked with Eva!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Changed to Configuring API Specification

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Thanks! Not applicable!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Removed as not needed now.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: There are several ways customers can use the expression. We cannot only keep only the recommended one.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Yes! Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Will let Eva know.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Not needed here.

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: I removed it anyhow. Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: I'm checking with Eva! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -1,61 +1,51 @@
...
 
-The policy validates the Client ID and, optionally, the Client Secret, of a client application created within an Anypoint Platform Organization.
+When a client application is registered in Anypoint Platform, a pair of credentials consisting of a client ID and client secret is generated. When the client application requests access to an API, a contract is created between the application and that API. An API that is protected with a Client ID Enforcement policy is accessible only to applications that have an approved contract.
 
-IMPORTANT: This policy will not validate successfully any client application credentials generated outside of the Anypoint Platform context.
+When you apply a Client ID Enforcement policy, access to your API is tracked by reporting the client ID along with the analytics events. You can configure Mule runtime engine (Mule) to encrypt all sensitive information that pertains to policies,contracts, and initialization data.
```
mtqp: I believe there's a missing space after the comma here: "(...)policies,contracts,(...)"

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -1,61 +1,51 @@
...
-NOTE: The set of Client ID and Client Secret are also known as credentials.
-
-= Analytics and  access tracking
-If you need to track client application access to your API, apply a policy that directly or indirectly makes sure that client ID is provided in the HTTP request (ie Client ID enforcement type policy).
-
-= Indirect client ID enforcement
-There are other provided policies that enforce client application credentials. For example:
+Default policies that internally enforce client application credentials include:
 
 * xref:rate-limiting-and-throttling-sla-based-policies.adoc[Rate Limiting - SLA-Based Policies]
```
mtqp, line 25: Beware that this may change after the unification you will do later on.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -1,61 +1,51 @@
...
 image::raml-snippet.png[]
 
-= Credentials source
+== Configuring How the Policy Obtains Credentials
 
-The policy can be configured to extract the client ID and client secret from the HTTP request in different ways using DataWeave 2.0 expressions.
+//You can configure the Client ID Enforcement policy to obtain the client credentials in one of two different ways: by using an HTTP basic authentication header or by using a custom expression.
 
-The following examples assume that the values for application credentials are '1234' for client ID and 'abcd' for client secret.
+You can configure the policy to extract either both the client ID and client secret, or only the client ID from the HTTP request by using a variety of custom DataWeave 2.0 expressions. The following examples use a client ID of '1234' and a client secret of 'abcd'.
```
mtqp, line 47: Using DataWeave 2.0 is only valid in Mule 4.0.0+
For 3.x Mules customer should use MEL expressions.

mtqp: You further detail this later on, feel free to close this if you think it is not necessary to put it here too.

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -1,61 +1,51 @@
...
+Before a client application is allowed to consume an API protected by a Client ID enforcement policy, the client must request access to the API. After an approved contract exists between the application and the API, every request must include the client credentials in compliance with how the policy is configured.
 
-= REST Console
-To prevent user requests from being rejected when using the REST Console, create a trait in the RAML root and then reference it in every operation of your API affected by policy. You may find the RAML snippet link containing the RAML code you need to add to the RAML in the corresponding policy from the list of applied policies in the Policies tab of your API specification in API Manager.
+For example, if the policy is configured to expect a client ID and client secret as headers, the application must send those credentials in the request, using the corresponding headers:
 
 image::raml-snippet.png[]
```
mtqp, line 41: Should this still go here or it has to be moved to the raml snippet section?

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Noted! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Ok! Thanks!

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: No this is fine here! Thanks!

evangelina (approved):
LGTM

PR #162, opened 2019-11-11
-n needed for the echo

jazahn:
A newline character is added by echo by default, `-n` strips that out. Without it, the newline gets encoded and the resulting string does not work.


glenn-rodgers-sf:

PR #169, opened 2019-11-19
DOCS-3692-made edits

priyaayyar:


modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
 endif::[]
```
kbostrom: I don't understand the title of this topic. 
Is this a Level-1 topic about a URL? 
That sounds pretty narrow for a major topic.

kbostrom: GLOBALLY replace the possessive ('s) with just the plain term, as an adjective: Policy Template Source URL (although I don't think that this is the title we want here...).

kbostrom: This current title implies that the following text is about the URL from which to source policy templates.
I don't think that this is what you intend, that the topic be simple a URL.
What is the user task, here, and WHY does your audience want or need the information in this topic?

kbostrom: I suggest that you change the topic title to something like "Download Policy Templates" or "Obtain the URL to Download Policy Templates."
Then I think the text should begin by explaining the circumstances in which your audience wants or needs to do this or to have this information: where is the user in all of this?

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 :keywords: policy, CloudFront
 
-[NOTE]
---
-The following section describes configurations parameters that are relevant for only Mule versions 4.1.7 and later, upto 4.2.0 and 4.2.2 and later, upto 4.3.0.
---
+The configuration parameters described in this section are relevant to only Mule runtime engine (Mule) versions 4.1.7 and later.
```
kbostrom: The topic should not begin with this text. The audience has no reference for "the parameters" yet.
The topic should begin by introducing what is covered here and WHY the audience wants or needs it.

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 
-When an application with a tracked flow is deployed to a Mule Runtime (Mule), the API Gateway of that Mule environment communicates with API Manager to request for the applied policies.
+When an application with a tracked flow is deployed to a Mule environment, the API gateway of that Mule environment communicates with API Manager to request for the applied policies.
```
kbostrom: Where is the user/audience? Is this accurate:
```suggestion
When you deploy an application with a tracked flow to a Mule environment, the API gateway of that Mule environment communicates with API Manager to request the policies applied to the environment.
```
Does the application require the policies or the policy templates for the environment? The next sentence refers to the latter; this sentence the former.

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 
-If policies are to be applied and the policy templates are not already present in Mule, API Gateway requests API Manager for the URLs from where to download the policy template.
+If the templates for the policies that are to be applied are not already present in Mule, API gateway requests API Manager for the URLs from where to download the template.
```
kbostrom: The previous sentence refers to applied policies; this one refers to templates for policies that are to be applied (not policies: templates, and not applied: to be applied). The two sentences are inconsistent, right at the start of the topic.

Are the policies applied or are there just templates for the policies to be applied, templates that the user has to configure? Again, I have no idea what the user is supposed to DO here, or why.

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 
-You can configure the following property to provide Policy Templates to API Manager only from a specified source:
+You can configure the following property to provide Policy templates to API Manager from a specified source:
```
kbostrom: Is this something that you do before or after the situation introduced in the previous two sentences?

kbostrom: This sentence says that you configure a property that obtains "from a specified source" a policy template that is then given TO API Manager...when the previous text states that the gateway attempts to get the policy template FROM API Manager. The flow for obtaining the template is unclear. What is the "specified source"? Is that the URL? Where does the information about the URL travel from, and where end?

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 
 `anypoint.platform.policy_template_source`
 
 [%header,cols="2*a"]
```
kbostrom: The syntax above does not include the parameters "external" and "internal" that follow in the table.
How are the syntax and the table related?

kbostrom: Note that we should not be using a table to provide a term/syntax and its definition; we should use a definition or description list, per our style.

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 |===
 |Value | Description
-|external | Indicates that the policy templates are to be retrieved from the legacy system, This is the default value to maintain compatibility with previously configured firewalls.
-|internal | Indicates that the policy templates are to be retrieved from the new system. This option provides additional reliability due to internal rerouting of the requests, when a server becomes unavailabile.
+|external | Policy templates to be retrieved from the legacy system. This is the default value to maintain compatibility with previously configured firewalls.
```
kbostrom: Does your audience know what the legacy system is, and why it is external?
Do we really need to mention the firewall? Isn't this irrelevant when the alternative is internal sourcing?

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
+|internal | Policy templates to be retrieved from the new system. This option provides additional reliability due to internal rerouting of the requests when a server becomes unavailable.
```
kbostrom: Are the templates actually "retrieved" from the new system or are they just there "within" the new system?
```suggestion
|internal | Policy templates are retrieved from within the new system. This option provides additional reliability due to internal rerouting of requests when a server becomes unavailable.
```

modules/ROOT/pages/policy-template-source.adoc
kbostrom:
```suggestion
|external | Policy templates are retrieved from the external legacy system. This is the default value to maintain compatibility with previously configured firewalls.
```

modules/ROOT/pages/policy-template-source.adoc
kbostrom: If these are the "parameters" referred to at the beginning, then this is the appropriate place to mention versioning restrictions.

modules/ROOT/pages/policy-template-source.adoc
```diff
@@ -4,27 +4,24 @@ include::_attributes.adoc[]
...
 |===
 
-If we start a v4.2.2 Mule Runtime, as shown in the box below, API Manager provides URLs for downloading the policy templates from Mule's CloudFront service.
+If you start Mule version 4.2.2 , as shown in the example, API Manager provides the URLs for downloading the policy templates from the Mule CloudFront service.
```
kbostrom: Does the example really show how the syntax configuration defined above works in practice?

kbostrom: Suggested rewording:
```suggestion
The following example shows a Mule 4.2.2 environment in which API Manager provides the URLs for downloading policy templates from the Mule CloudFront service:
```

kbostrom: Why is Mule CloudFront service not mentioned in the topic intro? Is it always the template source?

kbostrom (approved):
Hey Priya, I recommend that you have a peer writer review this topic, because I could not understand what it was trying to help the audience to understand or to do. 
Is this a Level-1 topic in your deliverable, or just for the PR? Because this seems like a really tiny issue (retrieving a template), and not something worthy of its own topic set.
I am approving this so that I do not hold you up, but I think that we have some work to do on this deliverable and its topic flow.

JaatadiaMulesoft (approved):

sarathecoubian (approved):

PR #177, opened 2019-12-10
Docs 6768 improve cors default policy pa

priyaayyar:


modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,47 @@ include::_attributes.adoc[]
...
 :keywords: cors, api gateway, gateway, policy
 
-Cross-Origin Resource Sharing (CORS) is a mechanism by which a web application can access resources defined in another domain. Browsers for example, rely on this standard for allowing cross origin requests.
+[%autowidth.spread,cols="a,a"]
+|===
+>s| Policy name | cross-origin resource sharing (CORS)
+>s| Summary      | Enables access to resources residing in external domains.
+>s| Category | Compliance
+>s| Available from Mule version | v4.1.1
+.4+>.^s| Returned Status Codes |
```
evangelina: The status code that returns are missing right? 

@mtqp could you fill it in? This is one of the confusing aspects about CORS.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,47 @@ include::_attributes.adoc[]
...
 
-The CORS policy complies with Cross-Origin Resource Sharing W3C Recommendation 16 January 2014. From API Manager, you can apply the policy, the default policy configuration considers the API to be a public resource. If not, removing the Public Resource check will enter the extended CORS configuration.
+CORS is a mechanism by which a Web application can access resources that are defined in another domain.  Using CORS, applications invoking JavaScript XMLHttpRequest (XHR) calls from a Web page can interact with resources from non-origin domains.
+
+Typically, Web browsers implement this standard for allowing cross origin requests. When you apply this policy to your API, it is considered a public resource by default. You can  check will enter the extended CORS configuration. If require that your API be shared with only specific resources, you can configure the policy accordingly.
```
evangelina: This sentence seems incomplete: "You can  check will enter the extended CORS configuration"

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,47 @@ include::_attributes.adoc[]
...
+
+The CORS policy complies with CORS W3C recommendation from 16 January 2014.
```
evangelina: Can we add a link to it?

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,47 @@ include::_attributes.adoc[]
...
+
+*<Eva: I saw that you can specify a default group in the UI. What is it? what does it do and what other groups can one create?>*
```
evangelina: @priyaayyar , please review this policy with @mtqp , we have been talking about this one and he can provide all the context.

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Will do!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Will do!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Noted, will work with Mariano!

modules/ROOT/pages/client-id-based-policies.adoc
```diff
@@ -32,7 +32,7 @@ Default policies that internally enforce client application credentials include:
 
 Although token enforcement policies can optionally skip client validation, it is recommended that you enforce the validation to ensure that the token is associated to an approved contract.
 
-== How the Policy Works
+== How This Policy Works
 
 Before a client application is allowed to consume an API protected by a Client ID enforcement policy, the client must request access to the API. After an approved contract exists between the application and the API, every request must include the client credentials in compliance with how the policy is configured.
```
nahuel-dallavecchia: "the client must request" ->"the client application must request"

nahuel-dallavecchia: " exists between the application and" -> " exists between the client application and"

nahuel-dallavecchia: "client credentials" -> "client application credentials"

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
+>s| Policy name | cross-origin resource sharing (CORS)
+>s| Summary      | Enables access to resources residing in external domains.
+>s| Category | Compliance
+>s| Available from Mule version | v4.1.1
+.4+>.^s| Returned Status Codes |
 
-The CORS policy complies with Cross-Origin Resource Sharing W3C Recommendation 16 January 2014. From API Manager, you can apply the policy, the default policy configuration considers the API to be a public resource. If not, removing the Public Resource check will enter the extended CORS configuration.
+CORS is a mechanism by which a Web application can access resources that are defined in another domain.  Using CORS, applications invoking JavaScript XMLHttpRequest (XHR) calls from a Web page can interact with resources from non-origin domains.
+
+Typically, Web browsers implement this standard for allowing cross origin requests. Your API is categorized as a public resource by default when you apply this policy. If you want to share your API with only specific resources, you can configure the policy as a restricted resource.
```
nahuel-dallavecchia: "If you want to share your API with only specific resource" -> "If you want to share specific resources of your API "

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
+
+The CORS policy complies with CORS W3C recommendation from 16 January 2014.
+
+*can we add a link to it?*
```
nahuel-dallavecchia: The reference document would be: https://www.w3.org/TR/cors/

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
+
+*<Eva: I saw that you can specify a default group in the UI. What is it? what does it do and what other groups can one create?>*
+
+You must specify the default group before specifying another one. The Default group is not a fallback in the normal sense of a default. In this case, it is only the first group you configure for CORS. -- Eva, not sure about this sentence. What does the fallbacl mean here. Please clarify.
```
nahuel-dallavecchia: In this context, fallback means that when there is an error (config error for example), instead of failing a general, it will use a default [and supposedly safe and proven] configuration.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
+
+You cannot order the CORS policy. This isn't a technical limitation, but rather by design. No OPTIONS request can reach the backend if CORS is configured, as this request is treated as a Preflight according to the CORS-compliant RFC.
+
+
+
+*Eva again the previous sentence does not make sense*
```
nahuel-dallavecchia: All policies has an order in which they are applied. Order matters because an incoming request will got through the policies in order before reaching the protected resource. 
HTTP protocol has many "verbs". Example: POST, GET, PUT, OPTIONS. If you have a protected request using OPTIONS, and the CORS policy is applied, then that particular request will never reach the backend (ie protected resource), since the policy itself will intercept *always* that request. When talking about CORS and OPTIONS, we are talking about PreFlight, which is a term used by the CORS specification (https://www.w3.org/TR/cors/#resource-preflight-requests) to identify which CORS operations are supported before performing then (operations may be costly in terms of allocated processing power).

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
 
 image::cors-policy.png[]
 
-NOTE: You must specify the default group before specifying another one. The Default group is not a fallback in the normal sense of a default. In this case, it is only the first group you configure for CORS.
+*Eva--do we have any prerequisites or scope and considerations for this policy?*
```
nahuel-dallavecchia: What do you mean by pre requisites or scope? If you mean some third party dependency as in the OAuth policies, then no. You can always apply this policy to a tracked API.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
 
-*Important:* The CORS policy cannot be ordered. This isn't a technical limitation, but rather by design. No OPTIONS request can reach the backend if CORS is configured, as this request is treated as a Preflight according to the CORS-compliant RFC.
+== How This Policy Works
 
-== CORS as an Interceptor
+*Eva we don't have much here.*
```
nahuel-dallavecchia: And we shouldn't I believe. This document is about the CORS policy, and the CORS interceptor, which the policy uses to do its magic, can be used in a standalone way, but that is out of the scope of this doc.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
 
-In Mule Runtime 4.0, the CORS Interceptor is a element in the HTTP Listener configuration. To be consistent, its configuration is the same as in the policy. Use this feature when the Mule Runtime has not been started with API Gateway capabilities and CORS functionality is needed.
+== Configuring CORS for non-API Gateway Mule Environments
 
-NOTE: Beware that in this scenario, altering the configuration will force an application redeploy.
+If your Mule runtime engine (Mule) is not started with API gateway capabilities and you need to implement the CORS functionality, you can use the CORS Interceptor, which is an element in the HTTP Listener configuration and is available in Mule 4.0.
```
nahuel-dallavecchia: API Gateway

nahuel-dallavecchia: Not sure: Should we replace "Mule runtime engine (Mule)" by "runtime"?

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,49 @@ include::_attributes.adoc[]
...
 
-The CORS configuration structure may change whether it is configured as a public resource or for a selected group of origins. The structure of the latter option is:
+When you modify the configuration, you must redeploy your application. The CORS configuration structure differs based on whether it is configured as a public resource or for a selected group of origins.
```
nahuel-dallavecchia: Incorrect semantic change. The original content stated that there are two main ways to leverage the policy capabilities: Public resource or Selected group of origins. No application redeploy involved. If you edit the policy configuration then, as with all other policies, policy is updated runtime side.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +86,51 @@ The public resource configuration looks like this:
...
-* Open a new Terminal window, paste the following line, and then press *Enter*: `open -a Google\ Chrome --args --disable-web-security`.
+When you apply the policy to your API from the UI, the following parameters are displayed:
 
-=== Google Chrome for Windows
+%header%autowidth.spread,cols="a,a,a"]
+|===
+| Element | Description | Required
+| Public resource | Whether the CORS configuration is to be applied as a public resource (default). | Yes
+| Default Group | This option is displayed when you uncheck the Public resource box. | No
+| Support credentials | Eva what are these?| No
```
nahuel-dallavecchia: Credentials, in the context of CORS spec, are cookies, authorization headers or TLS client certificates.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +86,51 @@ The public resource configuration looks like this:
...
 
-* Open a new Command Prompt window, navigate to the location of the Chrome executable (Chrome.exe), paste the following line, and then press *Enter*: `chrome.exe --disable-web-security`.
+|===
 
-=== Internet Explorer
+== Configuring API Specifications
 
-Enable the option to access data sources across domains. In some versions of IE, this option is in Internet Properties > Security tab > Custom Level in Security Level for this Zone > Miscellaneous > Enable Access to Data Sources Across Domains.
+*Eva-- anything to add here?*
```
nahuel-dallavecchia: What would you suggest?

nahuel-dallavecchia (changes requested):

modules/ROOT/pages/client-id-based-policies.adoc
priyaayyar: Done! Thanks!

priyaayyar: Done! Thanks!

priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Ok! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Ok! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: No gateway is correct. Done! Thanks!

priyaayyar: No again product naming is an editor thing. it is fine as is. Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: No idea. :-( Removed it!

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
 endif::[]
 :keywords: cors, api gateway, gateway, policy
 
-Cross-Origin Resource Sharing (CORS) is a mechanism by which a web application can access resources defined in another domain. Browsers for example, rely on this standard for allowing cross origin requests.
+[%autowidth.spread,cols="a,a"]
+|===
+>s| Policy name | cross-origin resource sharing (CORS)
+>s| Summary      | Enables access to resources residing in external domains.
```
kbostrom:
```suggestion
>s| Summary      | Enables access to resources residing in external domains
```
Not a sentence, but if there are others that ARE, then all get periods

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+>s| Category | Compliance
+>s| Available from Mule version | v4.1.1
+.4+>.^s| Returned Status Codes |
 
-The CORS policy complies with Cross-Origin Resource Sharing W3C Recommendation 16 January 2014. From API Manager, you can apply the policy, the default policy configuration considers the API to be a public resource. If not, removing the Public Resource check will enter the extended CORS configuration.
+CORS is a mechanism by which a Web application can access resources that are defined in another domain.  Using CORS, applications invoking JavaScript XMLHttpRequest (XHR) calls from a Web page can interact with resources from non-origin domains.
```
kbostrom: GLOBALLY lowercase "web." Same with "internet." They have become common nouns.
```suggestion
CORS is a mechanism by which a web application can access resources that are defined in another domain.  Using CORS, applications invoking JavaScript XMLHttpRequest (XHR) calls from a web page can interact with resources from non-origin domains.
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+
+Typically, Web browsers implement this standard for allowing cross origin requests. Your API is categorized as a public resource by default when you apply this policy. If you want to share only specific resources of your API, you can configure the policy as a restricted resource.
```
kbostrom:
```suggestion
Typically, web browsers implement this standard for allowing cross-origin requests. Your API is categorized as a public resource by default when you apply this policy. If you want to share only specific resources of your API, you can configure the policy as a restricted resource.
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+
+The CORS policy complies with https://www.w3.org/TR/cors/[CORS W3C recommendation] from 16 January 2014.
```
kbostrom:
```suggestion
The CORS policy complies with the https://www.w3.org/TR/cors/[CORS W3C recommendation] from 16 January 2014.
```
Are we saying that the compliance started 16 Jan or that the recommendation is dated 16 Jan? Reword to clarify.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+
+You must specify the default group before specifying another one. The Default group on this case means that it is the first group that you must configure for CORS. Therefore, when a configuration error occurs, the policy resets to the default configuration.
```
kbostrom:
```suggestion
You must specify the default group before specifying another. The default group, in this case, is the first group that you must configure for CORS. When a configuration error occurs, the policy resets to the default configuration.
```
Is the final sentence intended to emphasize that the config resets to the default config or to the first group configured? I removed "Therefore," because that would imply the latter; please reinsert it and reword to specify first group rather than default, if that is the case.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+
+By design, you cannot change the order of execution for the CORS policy. If a protected request using OPTIONS is sent to an application that has the CORS policy applied, the request will not reach the protected resource. According to the https://www.w3.org/TR/cors/#resource-preflight-requests[CORS specification], all OPTIONS requests are considered PreFlight.
```
kbostrom: Is it important to note "By design"? When using this phrase, the text typically tells WHY it was designed that way.

kbostrom: Will the audience understand why a preflight option is not accepted by a protected resource?

kbostrom: PreFlight > preflight 
I looked it up: https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+By design, you cannot change the order of execution for the CORS policy. If a protected request using OPTIONS is sent to an application that has the CORS policy applied, the request will not reach the protected resource. According to the https://www.w3.org/TR/cors/#resource-preflight-requests[CORS specification], all OPTIONS requests are considered PreFlight. 
 
 image::cors-policy.png[]
 
-NOTE: You must specify the default group before specifying another one. The Default group is not a fallback in the normal sense of a default. In this case, it is only the first group you configure for CORS.
 
-*Important:* The CORS policy cannot be ordered. This isn't a technical limitation, but rather by design. No OPTIONS request can reach the backend if CORS is configured, as this request is treated as a Preflight according to the CORS-compliant RFC.
+== How This Policy Works
+
+*Eva we don't have much here.*
```
kbostrom: This topic appears above.
Also, this text would end up in the publication, so be careful.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
 
-== CORS as an Interceptor
+== Configuring CORS for non-API Gateway Mule Environments
 
-In Mule Runtime 4.0, the CORS Interceptor is a element in the HTTP Listener configuration. To be consistent, its configuration is the same as in the policy. Use this feature when the Mule Runtime has not been started with API Gateway capabilities and CORS functionality is needed.
+If your Mule runtime engine (Mule) is not started with API gateway capabilities and you need to implement the CORS functionality, you can use the CORS Interceptor, which is an element in the HTTP Listener configuration and is available in Mule 4.0.
```
kbostrom: It doesn't bother me, but this is a long sentence. You might break it in two.

kbostrom: Do we want to say "is not started" or "is not configured"? Or...:
```suggestion
If your Mule runtime engine (Mule) is not enabled for API gateway and you need to implement the CORS functionality, you can use the CORS Interceptor. The CORS Interceptor is an element in the HTTP Listener configuration made available in Mule 4.0.
```
Because the main limitation to using Interceptor is you Mule version, you might reword to make that more prominent.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
 
-NOTE: Beware that in this scenario, altering the configuration will force an application redeploy.
+You can leverage the CORS policy capabilities as a public resource or as a selected group of origins, based on which the CORS configuration structure changes.
```
kbostrom: "based on which" made me read this sentence more than once, because I first thought that it referred to criteria for using it as a public resource vs. group of origins.
Are we really saying that if you leverage the policy as a selected group of origin, that this one option of the two changes the CORS config?

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -4,24 +4,40 @@ include::_attributes.adoc[]
...
+== Configuring CORS for non-API Gateway Mule Environments
 
-In Mule Runtime 4.0, the CORS Interceptor is a element in the HTTP Listener configuration. To be consistent, its configuration is the same as in the policy. Use this feature when the Mule Runtime has not been started with API Gateway capabilities and CORS functionality is needed.
+If your Mule runtime engine (Mule) is not started with API gateway capabilities and you need to implement the CORS functionality, you can use the CORS Interceptor, which is an element in the HTTP Listener configuration and is available in Mule 4.0.
 
-NOTE: Beware that in this scenario, altering the configuration will force an application redeploy.
+You can leverage the CORS policy capabilities as a public resource or as a selected group of origins, based on which the CORS configuration structure changes. 
 
-The CORS configuration structure may change whether it is configured as a public resource or for a selected group of origins. The structure of the latter option is:
+Selected Group of Origins Structure Example:
```
kbostrom: This should be either a subtopic, a list item, or a plain text introduction (sentence or fragment).
We want to avoid informal headings; also, as a heading, this would not end with a colon.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -48,8 +64,9 @@ The CORS configuration structure may change whether it is configured as a public
 </http:listener-interceptors>
 ----
 
-The public resource configuration looks like this:
+Public Resource Example:
```
kbostrom: DITTO previous comment.

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
 </http:listener-interceptors>
 ----
 
-If you cannot implement CORS for your API, another possible solution to an unreachable API is to disable the same-origin restrictions in your browser.
+== Disabling Same-Origin Restrictions in Your Browser
+
+If you cannot implement CORS for your API, you can disable the same-origin restrictions in your browser to call an inaccessible API. Each browser handles these restrictions in a unique way. You can research the tips, tricks, and plugins for your browser from the internet.
```
kbostrom: Consider rewording to clarify that what causes the problem is that CORS and API do not have the same origin. Is that the problem? Maybe something like "You might need to disable the same-origin restrictions in your browser if you cannot access an API that is not of the same origin as CORS."
UPDATED BELOW

kbostrom: However, if you reword (or even if you do not), the rest of the paragraph seems to focus on how browsers implement such restrictions, and not upon disabling them.
UPDATED BELOW

kbostrom: I also think it might be a little too casual to mention tips, tricks, and plugins for the browser on the internet. Maybe something more like "See documentation for your browser to determine how to disable the restriction"?
UPDATED BELOW

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+
+Before you modify your browser settings, make sure you understand and assess its potential security implications. You must use these browser settings only for testing in your own web pages. Otherwise, the browser can become vulnerable to malicious scripts and other potential threats.
```
kbostrom: Suggest omitting in favor of shorter mention above. What I really wanted to do was find a way to not get too far from the text that introduces the TASKS of disabling the restriction....

modules/ROOT/pages/cors-policy.adoc
kbostrom: UPDATE
AH! Now that I have read further, I understand that this is an intro to TASKS that disable the restriction in various browsers. So...how about this:
```suggestion
If you cannot implement CORS for your API, you might need to disable the same-origin restrictions in your browser to call an inaccessible API. How you modify your browser to do this depends on whether you use Chrome or Internet Explorer. However, before you modify your browser settings, be sure to consider the potential security implications and restrict changes to your own web pages to avoid external threats.
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+
+=== Disabling Same-Origin Restrictions on Google Chrome for Mac OS X
+
+To disable same-origin restrictions on Google Chrome for Mac OS X:
+
+* Open a new Terminal window and paste the following line: `open -a Google\ Chrome --args --disable-web-security`.
```
kbostrom: Paste from where? And don't they need to copy first?
Steps:
```suggestion
. Open a new Terminal window and paste within it the following line: 
`open -a Google\ Chrome --args --disable-web-security`
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+* Press *Enter*.
+
+=== Disabling Same-Origin Restrictions on Google Chrome for Windows
 
-== Disable Same-Origin Restrictions in your Browser
+To disable same-origin restrictions on Google Chrome for Windows:
 
-Each browser handles these restrictions in a unique way and tips, tricks, and plugins are available on the internet.
+* Open a command prompt window.
```
kbostrom: Is this the same as Terminal above, or are they named differently in Mac and Windows?

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+* Navigate to the location of the Chrome executable (Chrome.exe) file.
```
kbostrom:
```suggestion
. Navigate to the location of the Chrome executable file (Chrome.exe).
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+* Paste the following line: `chrome.exe --disable-web-security`.
```
kbostrom:
```suggestion
. Copy from Chrome.exe the following line and paste it in the command prompt window: 
`chrome.exe --disable-web-security`
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+If you cannot implement CORS for your API, you can disable the same-origin restrictions in your browser to call an inaccessible API. Each browser handles these restrictions in a unique way. You can research the tips, tricks, and plugins for your browser from the internet.
+
+Before you modify your browser settings, make sure you understand and assess its potential security implications. You must use these browser settings only for testing in your own web pages. Otherwise, the browser can become vulnerable to malicious scripts and other potential threats.
+
+=== Disabling Same-Origin Restrictions on Google Chrome for Mac OS X
+
+To disable same-origin restrictions on Google Chrome for Mac OS X:
+
+* Open a new Terminal window and paste the following line: `open -a Google\ Chrome --args --disable-web-security`.
+* Press *Enter*.
```
kbostrom:
```suggestion
. Press *Enter*.
```

modules/ROOT/pages/cors-policy.adoc
kbostrom: Steps below:

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+=== Disabling Same-Origin Restrictions on Google Chrome for Windows
 
-== Disable Same-Origin Restrictions in your Browser
+To disable same-origin restrictions on Google Chrome for Windows:
 
-Each browser handles these restrictions in a unique way and tips, tricks, and plugins are available on the internet.
+* Open a command prompt window.
+* Navigate to the location of the Chrome executable (Chrome.exe) file.
+* Paste the following line: `chrome.exe --disable-web-security`.
+* Press *Enter*.
```
kbostrom:
```suggestion
. Press *Enter*.
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
 ----
 
-If you cannot implement CORS for your API, another possible solution to an unreachable API is to disable the same-origin restrictions in your browser.
+== Disabling Same-Origin Restrictions in Your Browser
+
+If you cannot implement CORS for your API, you can disable the same-origin restrictions in your browser to call an inaccessible API. Each browser handles these restrictions in a unique way. You can research the tips, tricks, and plugins for your browser from the internet.
+
+Before you modify your browser settings, make sure you understand and assess its potential security implications. You must use these browser settings only for testing in your own web pages. Otherwise, the browser can become vulnerable to malicious scripts and other potential threats.
+
+=== Disabling Same-Origin Restrictions on Google Chrome for Mac OS X
```
kbostrom:
```suggestion
=== Disable Same-Origin Restrictions on Google Chrome for Mac OS X
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+
+To disable same-origin restrictions on Google Chrome for Mac OS X:
+
+* Open a new Terminal window and paste the following line: `open -a Google\ Chrome --args --disable-web-security`.
+* Press *Enter*.
+
+=== Disabling Same-Origin Restrictions on Google Chrome for Windows
```
kbostrom:
```suggestion
=== Disable Same-Origin Restrictions on Google Chrome for Windows
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+To disable same-origin restrictions on Google Chrome for Windows:
 
-Each browser handles these restrictions in a unique way and tips, tricks, and plugins are available on the internet.
+* Open a command prompt window.
+* Navigate to the location of the Chrome executable (Chrome.exe) file.
+* Paste the following line: `chrome.exe --disable-web-security`.
+* Press *Enter*.
 
-Make sure you understand the potential security implications of changing browser security settings. You should only use these options for testing on your own web pages because the browser can become vulnerable to malicious scripts and other potential threats.
+=== Disabling Same-Origin Restrictions on Internet Explorer
```
kbostrom:
```suggestion
=== Disable Same-Origin Restrictions on Internet Explorer
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
 
-=== Google Chrome for Mac OS X
+To disable same-origin restrictions on Internet Explorer:
 
-* Open a new Terminal window, paste the following line, and then press *Enter*: `open -a Google\ Chrome --args --disable-web-security`.
+* Navigate to *Internet Properties* > *Security* > *Custom Level in Security Level for this Zone* > *Miscellaneous* .
```
kbostrom:
```suggestion
. Navigate to *Internet Properties* > *Security* > *Custom Level in Security Level for this Zone* > *Miscellaneous*.
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+* Enable the Enable Access to Data Sources Across Domains option to access data sources across domains.
```
kbostrom: "Enable" how: select? click? 
```suggestion
. Enable the *Enable Access to Data Sources Across Domains* option to access data sources across domains.
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
-=== Google Chrome for Windows
+== Configuring Policy Parameters
 
-* Open a new Command Prompt window, navigate to the location of the Chrome executable (Chrome.exe), paste the following line, and then press *Enter*: `chrome.exe --disable-web-security`.
+When you apply the policy to your API from the UI, the following parameters are displayed:
 
-=== Internet Explorer
+%header%autowidth.spread,cols="a,a,a"]
+|===
+| Element | Description | Required
```
kbostrom:
```suggestion
| Element | Description | Required?
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+| Public resource | Whether the CORS configuration is to be applied as a public resource (default). | Yes
```
kbostrom:
```suggestion
| Public resource | Whether the CORS configuration is to be applied as a public resource (default) | Yes
```

modules/ROOT/pages/cors-policy.adoc
```diff
@@ -60,25 +77,47 @@ The public resource configuration looks like this:
...
+| Default Group | Whether the CORS configuration is to be applied only to specific resources. This option is displayed when you uncheck the Public resource box. | No
```
kbostrom:
```suggestion
| Default group | Whether the CORS configuration is to be applied only to specific resources (requires unselecting *Public resource*) | No
```

kbostrom (approved):
Done!

kbostrom (approved):
Done!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Yes, by design is required. The rest is fine as is too.

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Removed!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Changed to: The CORS configuration differs based on whether you leverage the CORS policy capabilities as a public resource or as a selected group of origins:

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/cors-policy.adoc
priyaayyar: Done! Thanks!

nahuel-dallavecchia (approved):

PR #189, opened 2020-01-16
AGW-3860: Mule 3 proxy elements.

nahuel-dallavecchia:


modules/ROOT/pages/proxy-mule3-elements.adoc
```diff
@@ -0,0 +1,24 @@
+= Mule 3.x API Proxy elements
+ifndef::env-site,env-github[]
+include::_attributes.adoc[]
+endif::[]
+
+== Automatically Filtering Internal Mule HTTP Headers
+
+To avoid internal Mule HTTP headers from being leaked to the backend implementation or from being inlcuded in the response, the following two filters are defined:
```
kbostrom: typo and passive voice
```suggestion
To avoid internal Mule HTTP headers from being leaked to the backend implementation or from being included in the response, use the appropriate one of the following two filters:
```
OR
```suggestion
To avoid internal Mule HTTP headers from being leaked to the backend implementation or from being included in the response, use one of the following two filters, respectively:
```
...because there are two situations to avoid and two solutions, and order is important.
Consider separating the situations into separate texts:

To avoid...backend implementation, use the following filter:
     <filter syntax>
To avoid...from being included in the response, use the following filter:
     <filter syntax>

kbostrom (approved):
Here you go!

modules/ROOT/pages/proxy-mule3-elements.adoc
priyaayyar: Done! Thanks!

priyaayyar (approved):

PR #195, opened 2020-01-23
Docs 6616 document mule header filters for 3x pa

priyaayyar:


modules/ROOT/nav.adoc
```diff
@@ -38,7 +38,7 @@
  *** xref:proxy-deploy-runtime-fabric.adoc[Deploy an API Proxy to Runtime Fabric]
  ** xref:download-proxy-task.adoc[Downloading an API Proxy]
  ** xref:proxy-domain-support.adoc[Automatic Domain Support for Mule 4 API Proxies]
- ** xref:proxy-mule3-elements.adoc[Mule 3.x API Proxy Elements]
```
nahuel-dallavecchia, line 41: I'd prefer Proxy Elements or something in the same spirit. Doing that, will allow to add additional descriptions of stuff used on Mule 3 proxies, like WSDL functions, for example.

modules/ROOT/pages/proxy-mule3-elements.adoc
```diff
@@ -1,19 +1,18 @@
-= Mule 3.x API Proxy elements
+= Filtering Internal Mule HTTP Headers in Mule 3.x API Proxy
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-== Automatically Filtering Internal Mule HTTP Headers
 
-To avoid internal Mule HTTP headers from being leaked to the backend implementation or from being included in the response, use one of the following two filters, as appropriate:
+To avoid internal Mule HTTP headers from being leaked to the backend implementation or from being included in the response, Anypoint Platform automatically implements the following:
```
nahuel-dallavecchia: They are not implemented by Anypoint Platform. I would say instead something like: API proxies by default filter them using the following components:

nahuel-dallavecchia (changes requested):

modules/ROOT/nav.adoc
priyaayyar: I think creating a bucket to include all content is not a great way to organize content. When we need to add new content, we can redetermine. One thing to keep in mind is that topic titles should give a glimpse into what the topic talks about. Descriptive and specific headings are a requirement for good documentation. Hope that makes sense! Thanks!

modules/ROOT/pages/proxy-mule3-elements.adoc
priyaayyar: Done! Thanks!

modules/ROOT/nav.adoc
nahuel-dallavecchia: Ok. Take also into account that renaming a doc each time that we need to add content not always is a good thing. Also, we already walked the path of having micro articles with not so good results in general.

modules/ROOT/pages/proxy-mule3-elements.adoc
nahuel-dallavecchia: As I said in my previous comment: "They are not implemented by Anypoint Platform".

nahuel-dallavecchia: I also would remove the use of the word "automatically" here. I would like to disallow the notion that somehow the "automatic" process can be disabled.

nahuel-dallavecchia (changes requested):
Please, review resolved conversations.

nahuel-dallavecchia (approved):

rddheath (approved):

PR #211, opened 2020-03-12
DOCS-7227-changed the title and edited the page

priyaayyar:


modules/ROOT/pages/import-api-task.adoc
```diff
@@ -1,13 +1,19 @@
-= Creating an API instance by importing on API Manager
+= Creating an API Instance by Importing to API Manager
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-You can develop an API for and export it from API Manager for future use or for sharing. The exported API can be imported using this procedure. To perform this task, you need to have a ZIP file on your file system containing an API object exported from API Manager.
+You might have developed an API and exported it from API Manager to share it across your organization or to use it at a later time. You can then import this API into API Manager when required.
```
kbostrom:
```suggestion
Imagine that you develop an API that you export from API Manager so that you can share it across your organization or use it later. You can still import this API back into API Manager, as required. 
```
I'm not sure that I understood this correctly. 
Is the point that we are trying to make that after you create an API in API Manager, you can export it to be used elsewhere, but you can also later retrieve it? 
I get confused at "or to use it at a later time"; what does that mean, that you export it and save it somewhere outside of API Manager? Why would you do that? Under what circumstances?
And then why would you have to return it to API Manager?
So, this topic isn't really about creating by import, but about returning from export?
Consider running this intro para by another writer and see what they have to offer about clarifying the wording, because I don't understand what you are trying to say.

modules/ROOT/pages/import-api-task.adoc
```diff
@@ -1,13 +1,19 @@
...
 include::_attributes.adoc[]
 endif::[]
 
-You can develop an API for and export it from API Manager for future use or for sharing. The exported API can be imported using this procedure. To perform this task, you need to have a ZIP file on your file system containing an API object exported from API Manager.
+You might have developed an API and exported it from API Manager to share it across your organization or to use it at a later time. You can then import this API into API Manager when required. 
 
-. After logging into Anypoint Platform, click API Manager.
-. In API Administration, select *Manager API > Import API from the Desktop*.
-. Browse to and select the ZIP file that contains the exported API instance to import.
+== Task Prerequisite
```
kbostrom: You should not use this subhead, because there is not a peer subhead for the actual task, so it looks like the task is part of the Task Prerequisite text.
```suggestion
```

modules/ROOT/pages/import-api-task.adoc
```diff
@@ -1,13 +1,19 @@
...
+
+Before you import an API to API Manager, you must have already exported the API instance ZIP file to you file system on your local machine or a server.
```
kbostrom: Is the audience RETURNING an API to API Manager, or importing a "foreign" API?
Where is the ZIP file being exported FROM?
Does this mean the same thing as "Before you import...to API Manager, you must import to your local file system"?
```suggestion
Before you can import an API to API Manager, you must have exported the API's ZIP file to your local file system or server.
```

modules/ROOT/pages/import-api-task.adoc
```diff
@@ -1,13 +1,19 @@
...
+
+To import the API instance to API Manager:
```
kbostrom: So this is the actual task, not a prereq., so it you keep the prereq title, then you need to insert a title like "Import an API Instance to API Manager."
However, why is this worded differently from the main title: CREATE an API by IMPORTING...?
Create and import are two different things.

modules/ROOT/pages/import-api-task.adoc
```diff
@@ -1,13 +1,19 @@
...
+
+. Go to *API Manager > API Administration*.
+. From the main page, select *Manage API > Import*.
```
kbostrom: What is the main page?
Why specify this, if this is the location reached in Step 1?
```suggestion
. Select *Manage API > Import*.
```

modules/ROOT/pages/import-api-task.adoc
```diff
@@ -1,13 +1,19 @@
...
+. In the file browser, select the API instance ZIP file to import.
```
kbostrom: How did the user get to a browser?
This term might be misleading.
Can you use only...:
```suggestion
. Select the ZIP file of the API that you want to import.
```

kbostrom (approved):
Done!

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Added it to the previous step explained.

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Fixed as I saw fit. Thanks!

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Done! Thanks

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Done! Thanks!

PR #217, opened 2020-03-20
DOCS-7142 made edits

priyaayyar:


modules/ROOT/pages/manage-versions-instances-concept.adoc
```diff
@@ -3,25 +3,24 @@ ifndef::env-site,env-github[]
...
+== Managing API Instances in API Manager
 
-In API Manager, you may also create, import, and deprecate API instances.
+You can create, import, and deprecate API instances using API Manager. You can use instances of an API in multiple environments. You can also have multiple instances of an API in the same environment:
 
-On the API administration dashboard, you expand the API version name to display the instances if there are any:
+//image::jsonplaceholder2.png[]
 
-image::jsonplaceholder2.png[]
+You can create one instance of an API that serves as a proxy and create another to directly manage as a basic endpoint. You can apply policies to the API based on how you have configured the endpint. For example, you can apply caching policies to the proxy, and throttling policies and security policies to the basic endpoint.
```
kbostrom: version?
```suggestion
You can create one instance of an API version that serves as a proxy and create another to manage directly as an endpoint. You can apply policies to the API based on how you configure that endpoint. For example, you can apply caching policies to the proxy, and throttling policies and security policies to the endpoint. 
```
The example is not an example of the sentence before it, which tells you that you can apply policies to the API based on how you configure the endpoint but does not mention applying policies to the API through the proxy.
It is unclear here which is the dominant/manager instance: the endpoint or the proxy.

modules/ROOT/pages/manage-versions-instances-concept.adoc
```diff
@@ -3,25 +3,24 @@ ifndef::env-site,env-github[]
...
 
-On the API administration dashboard, you expand the API version name to display the instances if there are any:
+//image::jsonplaceholder2.png[]
 
-image::jsonplaceholder2.png[]
+You can create one instance of an API that serves as a proxy and create another to directly manage as a basic endpoint. You can apply policies to the API based on how you have configured the endpint. For example, you can apply caching policies to the proxy, and throttling policies and security policies to the basic endpoint. 
 
-You can use instances of an API in multiple environments, and you can have multiple instances in the same environment. You can create one instance of an API that serves as a proxy. You can create another to manage directly as a basic endpoint. You can apply caching policies to the proxy, for example, and throttling policies and security policies to the basic endpoint. For more information about API instance management, see "About Basic API Management Tasks" link below.
+Additionally, you can promote an API to another environment.
```
kbostrom, line 20: Is this called "promoting"? Does that mean that you can share the API, create a new instance, or move the API entirely? Can you then change the shared/moved instance independently of the original?

modules/ROOT/pages/manage-versions-instances-concept.adoc
```diff
@@ -3,25 +3,24 @@ ifndef::env-site,env-github[]
...
+//(Promoting an API requires Manage APIs Configuration permission to be set for a user.)
 
-In addition to API instance management in API Manager, you can perform other API management tasks. You can promote an API to another environment and configure an API endpoint, for example. (Promoting an API requires Manage APIs Configuration permission to be set for a user.)
+== Managing API Versions in Exchange
 
-== API Version Management in Exchange
-
-In Exchange you may add, delete, and deprecate API versions, which are called Exchange assets. An API version in Exchange can have its own RAML file, API Portal, and URL. For more information about API version management in Exchange, see "Anypoint Exchange" documentation.
+You can add, delete, and deprecate API versions, also known as Exchange assets, in Anypoint Exchange (Exchange). An API version in Exchange can have its own RAML file, API Portal, and URL.
```
kbostrom:
```suggestion
You can add, delete, and deprecate API versions, also known as _Exchange assets_, in Anypoint Exchange (Exchange). An API version in Exchange can have its own RAML file, API portal, and URL. 
```
This text reads as if it were introductory information about versioning. Should it be moved up in the text chronology?

kbostrom (approved):
Done!

modules/ROOT/pages/manage-versions-instances-concept.adoc
priyaayyar: Fixed! Thanks!

modules/ROOT/pages/manage-versions-instances-concept.adoc
priyaayyar: No promote is OK. And promote means that you can move an API from one environment to another, for example, from testing to production.

modules/ROOT/pages/manage-versions-instances-concept.adoc
priyaayyar: No it is Ok where it is because it is about what you can do with an API version.

PR #250, opened 2020-05-08
DOCS-8008 made edits

priyaayyar:
Added comment entry


modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -3,11 +3,12 @@ ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
+//Nahuel please review the whole topic and suggest changes
```
nahuel-dallavecchia: Consider saying API endpoint instead of API instance endpoint (for consistency with other docs).

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -3,11 +3,12 @@ ifndef::env-site,env-github[]
...
 You can set up one of the following types of endpoint management:
```
nahuel-dallavecchia: API endpoint management?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -3,11 +3,12 @@ ifndef::env-site,env-github[]
...
 
 * Direct management of the endpoint: Basic endpoint
```
nahuel-dallavecchia: Perhaps a simpler title -> "Basic endpoint"
and then, a one line description of what is that.

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -3,11 +3,12 @@ ifndef::env-site,env-github[]
...
 +
-Set up a basic endpoint.
+Set up a basic endpoint.
```
nahuel-dallavecchia: Considering that it's already in the corresponding section, perhaps the title should be: Setup or Configuration?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -3,11 +3,12 @@ ifndef::env-site,env-github[]
...
 
+//Nahuel please review the whole topic and suggest changes
 You can set up one of the following types of endpoint management:
 
 * Direct management of the endpoint: Basic endpoint
 +
-Set up a basic endpoint.
+Set up a basic endpoint. 
 +
 * Management by proxy: Endpoint with proxy
```
nahuel-dallavecchia: Proxy endpoint?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -3,11 +3,12 @@ ifndef::env-site,env-github[]
...
 +
```
nahuel-dallavecchia, line 13: Following statement is incorrect. Basic endpoint is when you have a Mule application with autodiscovery correctly defined against your API. Proxy endpoint is when you either you have a Mule application where you cannot add the autodiscovery stuff or your API is implemented in a technology different than Mule. Policies only can be applied to Mule apps with autodiscovery, so in the latter case, you need an autogenerated proxy where policies can be applied and that it forwards correctly the request to the corresponding API (backend implementation).

nahuel-dallavecchia: "In API Manager, click the version, or an instance of the API if you have more than once defined."

nahuel-dallavecchia: "Add Consumer Endpoint" no consumer endpoint in the UI."Implementation URI" is the term that should be used instead (see UI).

nahuel-dallavecchia: "Above API Configuration, in Add Consumer Endpoint" -> Implementation URI is within that section (see UI).

nahuel-dallavecchia: In general I would say review screenshot and match what we have currently in the UI with the actual explanation.

nahuel-dallavecchia: Remove specific Mule agent version from doc (ie "Mule Agent 1.9.0").

nahuel-dallavecchia: "In *Implementation URI*, either enter the URL of the site that hosts the API or accept the default, a dummy URI." if Endpoint with Proxy is selected, the autogenerated proxy will never work in this way. If Basic endpoint selected, the implementation URI is informative. The actual link between API and API endpoint is done using autodisocvery. The same can be said about autogenerated proxies, the problem of using a fake implementation URI in this case is not about the link, but when that all requests handled by the proxy will fail when re sent to the fake impl URI.

nahuel-dallavecchia: IdP stands for Identity Provider (https://en.wikipedia.org/wiki/Identity_provider)

nahuel-dallavecchia: "In *Advanced Options*, optionally assign an instance label for the proxy, or accept the label, and set a response timeout." Label and timeout are optional settings (the latter only when Proxy endpoint is selected).

nahuel-dallavecchia: "For example, changing the path to `/users` makes the REST call to the service at the implementation URI `+http://jsonplaceholder.typicode.com/users+`." When you change the path of the proxy, makes the implementation endpoint to be called with what comes after the proxy path. For example:
Proxy path: /users
Impl URI: http://jsonplaceholder.typicode.com
Call proxy: http://<proxy host>:<proxy port>/users/hello -> Proxy will call impl endpoint using: http://jsonplaceholder.typicode.com/hello

nahuel-dallavecchia (changes requested):

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Endpoint with proxy is better I feel.

nahuel-dallavecchia: LGTM!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+= Configuring an API Endpoint in API Manager
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-You can set up one of the following types of endpoint management:
+You can set up one of the following types of API endpoint management:
 
-* Direct management of the endpoint: Basic endpoint
+* Basic endpoint
```
kbostrom:
```suggestion
* Basic API endpoint configuration
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 +
-Set up a basic endpoint.
+Configure and directly manage a basic endpoint. Use this option when you have a Mule runtime Engine (Mule) application with autodiscovery correctly defined against your API.
 +
-* Management by proxy: Endpoint with proxy
+* Endpoint with proxy
```
kbostrom:
```suggestion
* Proxy API endpoint configuration
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 endif::[]
 
-You can set up one of the following types of endpoint management:
+You can set up one of the following types of API endpoint management:
 
-* Direct management of the endpoint: Basic endpoint
+* Basic endpoint
 +
-Set up a basic endpoint.
+Configure and directly manage a basic endpoint. Use this option when you have a Mule runtime Engine (Mule) application with autodiscovery correctly defined against your API.
```
kbostrom:
```suggestion
Use this option to directly interact with a Mule application configured for autodiscovery.
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 +
-* Management by proxy: Endpoint with proxy
+* Endpoint with proxy
 +
-Set up a basic endpoint if you create your API outside API Manager. For example, if you create the API using Mule Runtime, you don't need a proxy.
+Use this option if you either have a Mule application for which cannot configure autodiscovery or if your API is implemented using a technology different than one used to develop Mule.
```
kbostrom:
```suggestion
Use this option to interact by proxy with a Mule application that cannot be automatically discovered, either because it is not configured for autodiscovery or because it uses a technology different from that of the API. In this case, the proxy must be both automatically discoverable and able to correctly forward application requests to the API. This is known as _backend implementation_.
```
BUT...do we really need "or because it uses a technology different from that of the API"? It sounds to me like the real problem is that the app cannot be automatically discovered. Is that CAUSED by the different technologies? If so, that should be specified as the cause rather than a separate issue.

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
-= Configuring an API Instance Endpoint on API Manager
+= Configuring an API Endpoint in API Manager
```
kbostrom, line 1: I read the intro text about the two types of API endpoint management several times and suggested several rewordings before I realized that they should be reworded to address the API endpoint config and not the application config. So the "management" is not really managing the endpoint but rather configuring the endpoint to interact with the app either directly or by proxy. 
The original text changes the title of "configure" to text about "manage," and that is misleading, I think.
So the following are my final edits, based on what I finally figured out and I hope they make the text immediately more clear for the audience.

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-You can set up one of the following types of endpoint management:
+You can set up one of the following types of API endpoint management:
```
kbostrom:
```suggestion
How you configure your API endpoint to manage application requests depends on the relationship between your Mule application and your API:
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+Configure and directly manage a basic endpoint. Use this option when you have a Mule runtime Engine (Mule) application with autodiscovery correctly defined against your API.
 +
-* Management by proxy: Endpoint with proxy
+* Endpoint with proxy
 +
-Set up a basic endpoint if you create your API outside API Manager. For example, if you create the API using Mule Runtime, you don't need a proxy.
+Use this option if you either have a Mule application for which cannot configure autodiscovery or if your API is implemented using a technology different than one used to develop Mule.
 
-You can save and later modify configuration of an API versions or instance.
+You can apply policies only to Mule applications that are configured with autodiscovery. Therefore, for the endpoints with proxy option, you need an auto-generated proxy to which you can apply policies and that it forwards correctly the request to the corresponding API (backend implementation).
```
kbostrom: See suggestion to merge this para into the para above it.

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
-. In API Manager, click the version, or an instance, of an API.
-+
-If you previously created an instance of the version, click the instance you want to reconfigure.
-+
-API Configuration appears.
-+
-image::apiconfig.png[height=321,width=321]
-+
-. Above API Configuration, in Add Consumer Endpoint, optionally enter the public-facing URL for the API. For example, enter `+http://product.mydomain.com+`.
+To configure an API endpoint:
```
kbostrom:
```suggestion
To configure your API endpoint:
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+
+. In API Manager, click the version, or an instance of the API.
```
kbostrom:
```suggestion
. In API Manager, click either the version number or instance of the API for which you want to configure an endpoint.
```
What does this look like? Is the user sometimes presented with version numbers and sometimes presented with some alternative ("instance")?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 +
-. In *Managing Type* at the bottom of the page, select B*asic Endpoint* or *Endpoint with Proxy*.
+.The diagram illustrates the inputs you must provide to create the API endpoint.
```
kbostrom:
```suggestion
.The screen capture shows how to create and configure an endpoint for the selected API.
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+image::create-instance-endpoint-midp.png[Create API Endpoint]
```
kbostrom:
```suggestion
image::create-instance-endpoint-midp.png[Create and configure an API endpoint]
```
We must be clear here: is the user CREATING or CONFIGURING an endpoint? If both, that should be made clear in both the topic title and the intro text.

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 +
-. Select the *Mule version* box if you want to manage APIs in Mule 4 or later.
-. In *Implementation URI*, either enter the URL of the site that hosts the API or accept the default, a dummy URI.
+<1> Add a tag for API endpoint.
```
kbostrom: Are these callouts supposed to be user actions or just identifiers? To make this one parallel with the following two:
```suggestion
<1> A tag for the API endpoint
```
However, what is a "tag"? Is this really a name or some other type of identifier?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+<2> The endpoint type: Basic or Proxy.
```
kbostrom:
```suggestion
<2> The endpoint type: basic or proxy
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+<3> The client provider if you have configured an external client provider.
```
kbostrom:
```suggestion
<3> The client provider, if you have configured an external client provider
```
Does this client provider need explaining?

modules/ROOT/pages/configure-api-task.adoc
kbostrom: Also...does this screencap show the entire process, including steps not yet described?
I would expect a screencap following Step 1 to illustrate Step 1, but it sounds like this illustrates the entire task.
Consider breaking up screencap into parts that illustrate each step, step by step.

modules/ROOT/pages/configure-api-task.adoc
kbostrom: Why is this called out first when it appears to an optional step...at the END of this task?

modules/ROOT/pages/configure-api-task.adoc
kbostrom: Also, if including callouts in one big illustration, why give numbers to optional tag and client provider, but not the major steps?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. Optionally, specify the public-facing URL for the API in the *Implementation URI* field, for example: `+http://product.mydomain.com+`.
```
kbostrom: I wonder if this should be part of or subordinate to the first step?
```suggestion
. Optionally, specify the public-facing URL for the API in the *Implementation URI* field, for example, `+http://product.mydomain.com+`.
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. Scroll to the bottom of the page to the *API Configuration* section.
```
kbostrom, line 26:
```suggestion
. Scroll down to the *API Configuration* section.
```

modules/ROOT/pages/configure-api-task.adoc
kbostrom: Is the app configured for autodiscovery, or the API?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. In the *Managing Type* field, select the endpoint type:
+
+* *Basic Endpoint*: select if you have an API with autodiscovery configured.
+* *Endpoint With Proxy*: select a proxy deployment target CloudHub or Hybrid installation. Hybrid is a product for managing servers.
```
kbostrom: Suggest either fully defining in intro OR here, but not both.
Note the intro definition for proxy version does not match this one.

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. In the *Managing Type* field, select the endpoint type:
+
+* *Basic Endpoint*: select if you have an API with autodiscovery configured.
+* *Endpoint With Proxy*: select a proxy deployment target CloudHub or Hybrid installation. Hybrid is a product for managing servers. 
 +
-Typically, you accept the default. The URI is not required, and often unknown, at this point.
-. From *Client provider*, select the IdP that you want to use to authenticate access to this API from registered client applications:
+You can accept the default Path `/` for the proxy endpoint or enter another path. 
+// For example, changing the path to `/users` makes the REST call to the service at the implementation URI `+http://jsonplaceholder.typicode.com/users+`.
+. Select the *Mule version* box if you want to manage APIs in Mule 4 or above.
```
kbostrom:
```suggestion
. Select the *Mule version* box if you want to manage APIs in Mule 4 or later.
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. If you are creating the endpoint with proxy, in the *Implementation URI* field, either specify the URL of the site that hosts the API.
 +
-image::create-instance-endpoint-midp.png[]
+Otherwise, you can skip this step.
```
kbostrom: ". If you are using the proxy configuration, in the *Implementation URI* field, specify the URL of the site that hosts the API."

You had "either" in the orig; did you mean "either accept the default or specify..."?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. From *Client provider*, select the identity provider (IdP) that you want to use to authenticate access to this API from registered client applications.
```
kbostrom:
```suggestion
. In *Client provider*, select the identity provider (IdP) that you want to use to authenticate access to this API from registered client applications.
```
Isn't this optional, per callout <3> above?
Are these steps really in the order in which a user would perform them? It sounds like the user has to scroll up and down the page?

modules/ROOT/pages/configure-api-task.adoc
kbostrom: Wait...using the *Implementation URI* field is mentioned again, later in this task. 
Is the user scrolling up and down?

modules/ROOT/pages/configure-api-task.adoc
kbostrom: Also...this field is also mentioned above; is the user scrolling up and down?

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
 +
 If no client providers are set up for your environment and organization, this field is not visible.
-* If you chose *Endpoint With Proxy*, select a proxy deployment target CloudHub or Hybrid installation.
-+
-Hybrid is a product for managing servers. You need to register a server with Mule Agent 1.9.0 to deploy the API to a Mule 3.8 or later hybrid target.
-+
-If you choose Endpoint With Proxy, also accept the default Path `/` for the proxy endpoint, or enter another path. For example, changing the path to `/users` makes the REST call to the service at the implementation URI `+http://jsonplaceholder.typicode.com/users+`.
-. In *Advanced Options*, optionally assign an instance label for the proxy, or accept the label, and set a response timeout.
-.. At the top of API Configuration section, click *ADD A TAG* to enter a tag name, such as Released, to signal the status of an API version to users.
+. If you selected the *Endpoint with proxy* option previously in this procedure, from the *Advanced Options*, optionally assign an instance label for the proxy, or accept the label, and set a response timeout.
```
kbostrom:
```suggestion
. If you selected *Endpoint With Proxy*, from *Advanced Options*, optionally accept the default or assign an instance label for the proxy, and set a response timeout.
```

modules/ROOT/pages/configure-api-task.adoc
```diff
@@ -1,51 +1,50 @@
...
+. Optionally, at the top of *API Configuration* section, click *ADD A TAG* to enter a tag name, for example: Released, to signal the status of an API version to users.
```
kbostrom:
```suggestion
. Optionally, scroll up *API Configuration* and click *ADD A TAG* to enter a tag name (for example,`Released,`) to signal the status of the API version.
```

kbostrom (approved):
Edited

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Ok Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: No, the users are show both. An instance is associated to an API version. 
I'm saying this now:In API Manager, select the instance of the API for which you want to configure an endpoint.

priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: That much detail not warranted. Other parts are very simple and evident. Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Good catch! This is a remnant from the previous work. I removed the step from here. Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Keep to the bottom of the page, because it is really that. It was very hard for me to find where this section was. Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Shortened it here. But need the info when the users need to make the decision.

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

nahuel-dallavecchia (approved):

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: API. No change needed here.

modules/ROOT/pages/configure-api-task.adoc
priyaayyar: Done! Thanks!

PR #252, opened 2020-05-13
Added instructions for RTF apps.

PabloCamino:
We got a case where customer needed assistance to do this in RTF and it was not covered in this document. Though customer could have inferred the need to use secure properties, having it here should make it clearer.


glenn-rodgers-sf:

PR #256, opened 2020-05-27
OutputType is not a valid attribute

derekXlin:
outputType="response" will cause deployment error as it is not a valid attribute


glenn-rodgers-sf:

PR #276, opened 2020-08-31
AGW-4345: Add persistence note to caching policy page.

JaatadiaMulesoft:


balbifm (dismissed):

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -87,6 +87,8 @@ If this box is checked the values stored in the cache persist after a runtime re
 
 By default this option is unchecked.
 
+*Note:* When you upgrade a Mule version that has the HTTP Caching policy configured to use persistent store, the policy tries to maintain the entries stored by the previous Mule version. However in a worst case scenario, the entries in the cache are invalidated and the cache is populated again as new requests arrive. This manipulation of the entries in the chache occurs automatically and is shielded from the user.
```
wwarner-salesforce:
```suggestion
*Note:* When you upgrade a Mule version that has the HTTP Caching policy configured to use the persistent store, the policy tries to maintain the entries stored by the previous Mule version. However, in a worst case scenario, the entries in the cache are invalidated and the cache is populated again as new requests arrive. This manipulation of the entries in the cache occurs automatically and is shielded from the user.
```

3 suggested changes:
1. "use persistent store" -> "use the persistent store" (or maybe "use persistent storage"?)
2. comma after "However"
3. chache -> cache

wwarner-salesforce (dismissed):

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

wwarner-salesforce (approved):

PR #280, opened 2020-09-09
Include the latest version of RTF content.

fermujica:


modules/ROOT/pages/proxy-deploy-runtime-fabric.adoc
```diff
@@ -3,7 +3,7 @@ ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-include::1.0@runtime-fabric:ROOT:page$proxy-deploy-runtime-fabric.adoc[tag=content]
+include::runtime-fabric:ROOT:page$proxy-deploy-runtime-fabric.adoc[tag=content]
```
sanderson-sfdc, line 6: This works without a version?  Interesting!

sanderson-sfdc (approved):
One comment, but I'm sure you tested it.

modules/ROOT/pages/proxy-deploy-runtime-fabric.adoc
fermujica: It does. When no version is specified, it automatically uses the latest available.

PR #281, opened 2020-09-10
Docs 3716 update delete api task pa

priyaayyar:


modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,20 @@
-= Removing an API instance on API Manager
+= Remove API Instances
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-The procedure removes the capability to manage an API Version from API Manager. If the API resides in Exchange or on the file system as a ZIP, it isn't deleted from Exchange or the file system.
+You can remove a an API instance or a version of an API instance, if you no longer need to manage it from API Manager. when you remove the API version from API Manager, the API is not deleted from Anypoint Exchange (Exchange) or the file system as a ZIP, if it exists in any of these locations.
```
evangelina: * I think the "when" should start with uppercase. 
* Is it ok to name it Anypoint Exchange (Exchange)? sounds redundant
* I would remove the "as a ZIP"

sbehrends: I agree that "Anypoint Exchange (Exchange)" sounds redundant. Anypoint Exchange should be enough.

Also, regarding "the file system as a ZIP". I would specify that the API references to an API Asset in Exchange, and such asset won't be deleted.

sbehrends (changes requested):
Left some comments regarding removing an API Instance will not delete the Asset in Exchange.

modules/ROOT/pages/delete-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/nav.adoc
```diff
@@ -3,16 +3,16 @@
...
- ** xref:import-api-task.adoc[Creating an API Instance by Importing on API Manager]
- ** xref:export-api-latest-task.adoc[Exporting an API Instance on API Manager]
- ** xref:configure-api-task.adoc[Configuring an API Instance Endpoint on API Manager]
- ** xref:deprecate-api-latest-task.adoc[Deprecating an API Instance on API Manager]
- ** xref:delete-api-task.adoc[Removing an API Instance on API Manager]
+ ** xref:import-api-task.adoc[Creating an API Instance by Importing]
+ ** xref:export-api-latest-task.adoc[Exporting an API Instance]
+ ** xref:configure-api-task.adoc[Configuring an API Instance Endpoint]
+ ** xref:deprecate-api-latest-task.adoc[Deprecating an API Instance]
+ ** xref:delete-api-task.adoc[Removing an API Instance]
```
kbostrom: These titles all need updating to imperative voice, right?

modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,22 @@
...
+You can remove an API instance or a version of an API instance if you no longer need to manage it from API Manager. When you remove the API version from API Manager, the API is not deleted from Anypoint Exchange (Exchange) or from the file system, if it exists in any of these locations. 
 
-. In Anypoint Platform, click API Manager.
-. In API Administration, click the API version link or an instance link.
-. Select Delete from the Actions dropdown.
+Furthermore, the API asset in Exchange that the API references is not deleted.
+
+[Note] 
+--
+You cannot delete an API specification that is currently used in one or more API instances or versions being managed in API Manager.
```
kbostrom: I think that this is overly complex, referring to both an API instance and a version of an API instance (they are both instances) and then saying that it is not deleted elsewhere and also the Note about deleting "an API specification."
The topic refers to instances, not specifications.
Here is my best guess:
If you no longer need to manage an API instance in API Manager, you can remove it to <what is the reason for doing this?>. Removing the instance from API Manager does not remove it or assets that it references from any other locations.

modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,22 @@
...
-. In API Administration, click the API version link or an instance link.
-. Select Delete from the Actions dropdown.
+Furthermore, the API asset in Exchange that the API references is not deleted.
+
+[Note] 
+--
+You cannot delete an API specification that is currently used in one or more API instances or versions being managed in API Manager. 
+--
+
+To remove API instances or API versions from API Manager:
```
kbostrom: This is different from what was said above, which paired instances and version instances. 
Clarify: are these steps intended to remove one instance or a whole version and all of its instances?
Could we simple say:
```suggestion
To remove an API from API Manager:
```

modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,22 @@
...
+
+. Go to *Anypoint Platform* > *API Manager*.
+. In *API Administration*, click the *API version* link or an instance link.
```
kbostrom: Clarify this step; I suggest including a screen capture that shows what "an instance link" might look like.
Would the UI have both the version and instance links, or just one or the other?
Does clicking either result in the same UI from which the audience completes the following step?

modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,22 @@
...
+. Select *Delete* from the *Actions* dropdown.
```
kbostrom:
```suggestion
. Select *Delete* from the *Actions* menu.
```

modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,22 @@
...
+. Click *Delete* again to confirm the delete action.
```
kbostrom:
```suggestion
. Click *Delete* again to confirm. 
```

modules/ROOT/pages/delete-api-task.adoc
```diff
@@ -1,12 +1,22 @@
...
+To remove API instances or API versions from API Manager:
+
+. Go to *Anypoint Platform* > *API Manager*.
+. In *API Administration*, click the *API version* link or an instance link.
+. Select *Delete* from the *Actions* dropdown.
+. Click *Delete* again to confirm the delete action. 
 +
-. Respond to the prompt to confirm the deletion. Click Delete.
\ No newline at end of file
+Your API instance is now successfully removed.
```
kbostrom: GLOBALs:

Choose between "delete" and "remove." Because the UI uses the former, I suggest replacing all "remove"s with "delete"s, unless you clarify a difference.

Clarify what is being deleted: an API instance? a version instance? a version? The wording above is confusing because it uses these alternatively and seemingly interchangeably. Provide a REASON for doing this. Do not mention deleting a specification if it is separate from instance/version; that is off topic. Ditto "asset": why would an asset be removed?

kbostrom (dismissed):
Edited per scope of request.

modules/ROOT/nav.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/delete-api-task.adoc
priyaayyar: API specification is the definition of the API which is different from the API itself or instance or version. I have tried to streamline it based on your comment. Done! Thanks!

modules/ROOT/pages/delete-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/delete-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/delete-api-task.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/delete-api-task.adoc
priyaayyar: This is a very common thing, but I've added a screenshot as well. Done! Thanks!

sbehrends (approved):
Thanks for applying the feedback @priyaayyar.
LGTM 🚀

kbostrom (approved):
Reapproved without reedit

PR #293, opened 2020-10-19
Docs 9928 update rate limiting policy pa

priyaayyar:


mtqp (dismissed):

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
+= Rate Limiting Policy
+ifndef::env-site,env-github[]
+include::_attributes.adoc[]
+endif::[]
+:keywords: rate limiting, api gateway, gateway, policy
+
+[%autowidth.spread,cols="a,a"]
+|===
+>s| Policy name | Rate limiting
```
kbostrom: Policy name init. capped in the narrative, so I would do so here as well; please check other topics with this table cell?
```suggestion
>s| Policy name | Rate Limiting
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+>s| Summary      | Monitors access to an API by defining a maximum amount of requests processed for a period of time
```
kbostrom: See style ref., “amount,” “number”

```suggestion
>s| Summary      | Monitors access to an API by defining a maximum number of requests processed for a period of time
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+>s| Category | Quality of Service
+>s| First Mule version available | v4.1.0
+.4+>.^s| Returned Status Codes
+|400 - Quota exceeded by WSDL APIs that use SOAP v1.2. Requests are blocked until the current window completes
```
kbostrom: Review style reference entry for "completes" and GLOBALLY reword to fix. Here is just one suggestion:
```suggestion
|400 - Quota exceeded by WSDL APIs that use SOAP v1.2. Requests are blocked until the current window completes its processing.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+|429 - Quota exceeded, requests are blocked until the current window finishes
+|500 - Quota exceeded by WSDL APIs that use SOAP v1.1. Requests are blocked until the current window completes.
```
kbostrom: ditto edit for value 400

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+|===
+
+The Rate Limiting policy enables you to control the incoming traffic to an API by limiting the number of requests that the API can receive within a given period of time. After the limit is reached, the policy rejects all requests, thereby avoiding any additional load on the backend API.
+
+When you configure the Rate Limiting policy, you can specify any number or pairs of quota (number of requests) and time window (time period).
```
kbostrom: What does "any number or pairs" mean? Do you mean "any number OF pairs"?
Consider clarifying by rewording...perhaps:
```suggestion
When you configure the Rate Limiting policy, you can specify any number of pairs of quota (number of requests) and time window (time period) values. 
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+|429 - Quota exceeded, requests are blocked until the current window finishes
+|500 - Quota exceeded by WSDL APIs that use SOAP v1.1. Requests are blocked until the current window completes.
+
+|===
+
+The Rate Limiting policy enables you to control the incoming traffic to an API by limiting the number of requests that the API can receive within a given period of time. After the limit is reached, the policy rejects all requests, thereby avoiding any additional load on the backend API.
+
+When you configure the Rate Limiting policy, you can specify any number or pairs of quota (number of requests) and time window (time period). 
+
+Additionally, you can further configure the policy to run in a Mule cluster, or add identifiers to define groups of requests. In a clusterized policy configuration, the quota is shared between all of the nodes in the cluster. For more information about how these options work, see Examples.
```
kbostrom: "clusterized" or "clustered"? Emphasis on the fact that the config is of a cluster type or was modified to include clusters?

kbostrom: "shared between all" > "shared among all"

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+== How This Policy Works
+
+The Rate Limiting policy keeps track of the number of requests made in the current window (the available quota), allowing the requests to reach the backend only if the available quota is greater than zero.
```
kbostrom: "keep track of" is idiomatic; replace with "monitor":
```suggestion
The Rate Limiting policy monitors the number of requests made in the current window (the available quota), allowing the requests to reach the backend only if the available quota is greater than zero. 
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+When you configure the Rate Limiting policy, you can specify any number or pairs of quota (number of requests) and time window (time period). 
+
+Additionally, you can further configure the policy to run in a Mule cluster, or add identifiers to define groups of requests. In a clusterized policy configuration, the quota is shared between all of the nodes in the cluster. For more information about how these options work, see Examples.
+
+== How This Policy Works
+
+The Rate Limiting policy keeps track of the number of requests made in the current window (the available quota), allowing the requests to reach the backend only if the available quota is greater than zero. 
+
+You can configure the policy for multiple groups of requests by using `identifiers` in the policy configuration. Each group has a separate available quota for their window.
```
kbostrom: What do you intend to imply by putting "identifiers" in single quotation marks?
We reserve those for monospace formatting, for syntax input or output. If that is what you intend, then STET; if not, use double quotation marks.

kbostrom: "...each group...for its window": numerical agreement

```suggestion
You can configure the policy for multiple groups of requests by using identifiers in the policy configuration. Each group has a separate available quota for its window. 
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+Additionally, you can further configure the policy to run in a Mule cluster, or add identifiers to define groups of requests. In a clusterized policy configuration, the quota is shared between all of the nodes in the cluster. For more information about how these options work, see Examples.
+
+== How This Policy Works
+
+The Rate Limiting policy keeps track of the number of requests made in the current window (the available quota), allowing the requests to reach the backend only if the available quota is greater than zero. 
+
+You can configure the policy for multiple groups of requests by using `identifiers` in the policy configuration. Each group has a separate available quota for their window. 
+
+To understand how the Rate Limiting policy works, let’s take an example of how the configuration of 3 requests every 10 seconds allows or restricts the request, based on the quota available in that window:
```
kbostrom: NEVER use "let's" or "let us" or other wording that refers to "us." Address the audience directly, omitting any conversational tone that implies "we":
```suggestion
To understand how the Rate Limiting policy works, consider an example in which the configuration of 3 requests every 10 seconds allows or restricts incoming requests, based on the quota available in that window:
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+image::rate-limiting-generic-example.png[Rate Limiting Policy Example,70%,85%]
+
+In the first window, because the quota is reached with the third request, all subsequent requests are rejected until the window closes. In the second window, only 2 of the 3 requests are processed. In this window, the quota remaining for that window is dropped after the window time has elapsed.
```
kbostrom: Spell amounts under 10, unless a sentence contains values both under and over, as in line 32 earlier:
```suggestion
In the first window, because the quota is reached with the third request, all subsequent requests are rejected until the window closes. In the second window, only two of the three requests are processed and the quota remaining is dropped after the window time has elapsed.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+An accepted request passes through the API to the backend. A rejected request on the other hand, displays a “429 status for HTTP, ” (or either 400 or 500 if the API is WSDL) and does not reach the backend.
```
kbostrom: Never use the idiom "on the other hand"; replace with something like "however" or "conversely," and include appropriate punctuation. Also, remember that displayed syntax should be formatted using ` `:
```suggestion
An accepted request passes through the API to the backend. Conversely, a rejected request displays 1429 status for HTTP` (or either 400 or 500 if the API is WSDL) and does not reach the backend.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+== Examples
+
+Let’s look how the same configuration of 3 requests every 10 seconds works when the configuration is clusterized and identifiers are added.
```
kbostrom: ditto re. "let's":
```suggestion
Consider how the previously described configuration of 3 requests every 10 seconds works when the configuration is clusterized and identifiers are added.
```
BUT...does this sentence introduce ALL of the following Level-3 (===) examples, or just the first one? If the latter, rework this intro to address all examples, and then individually introduce each example within its own topic.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+=== Configuring Identifiers with Regular String
```
kbostrom: What does "with Regular String" mean? 
Configuring identifiers that have **a** regular string, or regular string**s**?
Or configuring identifiers **by using** a regular string or regular strings?
("regular string" needs an article or to be made plural.)

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+When you add identifiers to the policy configuration from the UI, you can define groups of requests. The configured limits apply independently for each group. You can also use the identifier `#[attributes.method]` for one bucket per HTTP method in your Rate Limiting policy configuration.
```
kbostrom: "when you do this to that from here" construction is awkward. Suggest...:
```suggestion
When you use the UI to add identifiers to the policy configuration, you can define groups of requests. The configured limits apply independently to each group. You can also use the identifier `#[attributes.method]` for one bucket per HTTP method in your Rate Limiting policy configuration. 
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+== Examples
+
+Let’s look how the same configuration of 3 requests every 10 seconds works when the configuration is clusterized and identifiers are added.
+
+=== Configuring Identifiers with Regular String
+
+When you add identifiers to the policy configuration from the UI, you can define groups of requests. The configured limits apply independently for each group. You can also use the identifier `#[attributes.method]` for one bucket per HTTP method in your Rate Limiting policy configuration. 
+
+//If the universe of possible requests can be  split into a finite number of groups, then each group can be assigned with a selector key. that resolves to a selector key (a String). One Rate Limit algorithm will be created per key.
```
kbostrom: Is this temporarily omitted text? (//)
If not...:
```suggestion
//If all possible requests can be split into a finite number of groups, then each group can be assigned a selector key (a String). One Rate Limit algorithm is created per key.  
```
Also, should be reworded to active voice rather than passive.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+== Examples
+
+Let’s look how the same configuration of 3 requests every 10 seconds works when the configuration is clusterized and identifiers are added.
+
+=== Configuring Identifiers with Regular String
+
+When you add identifiers to the policy configuration from the UI, you can define groups of requests. The configured limits apply independently for each group. You can also use the identifier `#[attributes.method]` for one bucket per HTTP method in your Rate Limiting policy configuration. 
+
+//If the universe of possible requests can be  split into a finite number of groups, then each group can be assigned with a selector key. that resolves to a selector key (a String). One Rate Limit algorithm will be created per key.  
+
```
kbostrom, line 51: GLOBALLY standardize how you present "identifier" and clarify what you mean: just a generic thing that identifies, or something seen in the UI (`identifier`) or some sort of proper noun/name (Identifier module).
I will stop asking about this; please search for and clarify throughout.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+An `Identifier` is a non-obligatory parameter.  By default the Identifier is blank. Based on how you configure the Identifier, the policy performs in the following ways:
```
kbostrom: "blank" or "has no value"?
If referring to "the `Identifier` field, that it is "empty" by default, or "left empty" or "left blank" or "has no value" by default.

kbostrom: Also, "Based on how you configure..." is not accurate; that implies different configurations rather than configured or NOT configured:
```suggestion
`Identifier` is a non-obligatory parameter.  By default, the 'Identifier' field is blank. Based on whether you accept the default or provide a value, the policy performs in the following ways:
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+* When not configured, the Rate Limiting quota applies to every request per bucket or group.
```
kbostrom:
```suggestion
* Not configured (default)
+
The Rate Limiting quota applies to every request per bucket or group.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* When configured for an obligatory HTTP header, each case-sensitive value of the header has its own quota.
```
kbostrom:
```suggestion
* Configured for an obligatory HTTP header
+
Each header has its own quota. Header values are case-sensitive.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+When you add identifiers to the policy configuration from the UI, you can define groups of requests. The configured limits apply independently for each group. You can also use the identifier `#[attributes.method]` for one bucket per HTTP method in your Rate Limiting policy configuration. 
+
+//If the universe of possible requests can be  split into a finite number of groups, then each group can be assigned with a selector key. that resolves to a selector key (a String). One Rate Limit algorithm will be created per key.  
+
+An `Identifier` is a non-obligatory parameter.  By default the Identifier is blank. Based on how you configure the Identifier, the policy performs in the following ways:
+
+* When not configured, the Rate Limiting quota applies to every request per bucket or group.
+* When configured for an obligatory HTTP header, each case-sensitive value of the header has its own quota. 
++
+Quotas are created using the _lazy creation_ strategy.
```
kbostrom: passive

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* When configured  for a non-obligatory HTTP header, custom header, payload, query parameter or expression, each value has its own quota.
```
kbostrom:
```suggestion
* Configured for a non-obligatory HTTP header
+
Custom header, payload, query parameter, or expression values each have their own quotas. 
```

modules/ROOT/pages/rate-limiting.adoc
kbostrom: Edits represent attempt to provide three parallel list entries. 
You can do the same thing in your own way, as you like.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+An `Identifier` is a non-obligatory parameter.  By default the Identifier is blank. Based on how you configure the Identifier, the policy performs in the following ways:
+
+* When not configured, the Rate Limiting quota applies to every request per bucket or group.
+* When configured for an obligatory HTTP header, each case-sensitive value of the header has its own quota. 
++
+Quotas are created using the _lazy creation_ strategy.
+* When configured  for a non-obligatory HTTP header, custom header, payload, query parameter or expression, each value has its own quota. 
++
+If the Identifier is not sent in the request, the Identifier defaults to an empty value, having its own quota. This behavior allows the Rate Limiting policy to be applied to an API consumed by uncontrolled clients, and at the same time accommodates special buckets for the clients sending the Identifier (which may be internal controlled services).
```
kbostrom: This sentence does not make sense, because it uses one term (the Identifier) to refer to two different things: the thing not included in the request and the value assigned to the thing that is included in the config options:

If the Identifier is not sent in the request, [this same Identifier] the Identifier defaults to an empty value, having its own quota.

kbostrom: To what does "which may be internal controlled services" refer? the buckets? the content of the buckets? something else? Clarify.

Also, "may" > "might"

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+The following example shows the order of events that occur over a period of time using the Identifier `#[attributes.method]` for a limit of 3 requests every 10 seconds:
+
+image::rate-limiting-configure-identifiers-example.png[Rate Limiting Configure Identifiers Example,70%,85%]
+In the example:
+
+* Every HTTP method is allowed 3 requests every 10 seconds (in this example only GET and POST requests have been made to the API).
```
kbostrom:
```suggestion
* Every HTTP method is allowed 3 requests every 10 seconds (in this example, only GET and POST requests are made to the API).
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* The Rate Limiting policy works in a fixed window fashion (see the window size bracket).
```
kbostrom:
```suggestion
* The Rate Limiting policy works in a fixed-window fashion (see the window size bracket).
```
See what window size bracket, and how does this explain "fixed-window fashion"?

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* The window start times are independent.
+* The engine uses a _lazy creation_ strategy that spools a rate limiting algorithm whenever the first request for a method is received.
```
kbostrom:
```suggestion
* The engine uses a _lazy creation_ strategy that spools a rate-limiting algorithm whenever the first request for a method is received.
```
Note also that "lazy creation" is used earlier but not defined there.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+=== Configuring Identifiers with DataWeave Expressions
+
+The rate limiting engine, which is HTTP agnostic, depends solely on the resolution of the DataWeave expression. You can alter the identifier expression to cover complex rate limiting scenarios.
```
kbostrom:
```suggestion
The rate-limiting engine, which is HTTP agnostic, depends solely on the resolution of the DataWeave expression. You can alter the identifier expression to cover complex rate-limiting scenarios.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+For example, you can configure a Rate Limiting policy with an Identifier that uses one bucket for all class A & C LAN requests, and another bucket for everything else. The following image illustrates the latter case, which corresponds to 3 requests per 10 seconds quota with the DataWeave expression #[attributes.queryParam[‘customIdentifier’]] as the policy identifier:
```
kbostrom: Clarify "...all class A & C LAN requests".
Do you mean all Class A and Class C LAN requests?
"all Class A&C LAN requests", where A&C is a single type of LAN request?

kbostrom: Does "...The following image illustrates the latter case" refer to the second bucket in the previous sentence?
Because the previous sentence sounds like ONE case in which you have TWO buckets, not two cases....

kbostrom: Copyedits only:
```suggestion
For example, you can configure a Rate Limiting policy with an Identifier that uses one bucket for all class A & C LAN requests and another bucket for everything else. The following image illustrates the latter case, which corresponds to a 3 requests per 10 seconds quota with the DataWeave expression #[attributes.queryParam[‘customIdentifier’]] as the policy identifier:
```

kbostrom: BREAK HERE

kbostrom (approved):
First half of CU edit, per scope of request

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+image::rate-limiting-configure-identifiers-dataweave.png[Rate Limiting Configure Identifiers with DataWeave,70%,85%]
+
+In the example:
+
+* All requests without the Identifier are resolved to the empty Identifier, and therefore use a single rate limiting algorithm.
```
kbostrom: ditto prev. question about 1 or 2 identifiers here. Also no need for comma...:
```suggestion
* All requests without the Identifier are resolved to the empty Identifier and therefore use a single rate limiting algorithm.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* Each different Identifier uses a different bucket, with its own independent quota.
```
kbostrom:
```suggestion
* Each different Identifier uses a different bucket, each bucket with its own independent quota.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+This configuration creates a false or a true bucket that corresponds to the locality of the IP that made the request. The False and True values correspond to the domain of boolean values and not HTTP.
```
kbostrom: Inconsistent capitalization of true and false.
If the second are UI display or syntax, use proper tagging.

kbostrom: We have different capitalizations for Boolean, based on usage. Which of these applies?

Boolean (non-object-type specification use)

Boolean class

boolean primitive

boolean primitive object type 

Boolean wrapper-class object type

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+image::rate-limiting-configure-identifiers-dataweave.png[Rate Limiting Configure Identifiers with DataWeave,70%,85%]
+
+In the example:
+
+* All requests without the Identifier are resolved to the empty Identifier, and therefore use a single rate limiting algorithm.
+* Each different Identifier uses a different bucket, with its own independent quota.
+
+This configuration creates a false or a true bucket that corresponds to the locality of the IP that made the request. The False and True values correspond to the domain of boolean values and not HTTP. 
+
```
kbostrom, line 92: GLOBAL and similarly, string or String?

string, String type

String class

string object (Internet search for all string terms)

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+Nevertheless, the policy works correctly because the engine treats the resolved expression as a String. In this case, the value is automatically cast  from Boolean to String. You can explicitly define casting in DataWeave by adding “output text/plain ---“ to your script.
```
kbostrom:
```suggestion
Nevertheless, the policy works correctly because the engine treats the resolved expression as a String. In this case, the value is automatically cast from Boolean to String. You can explicitly define casting in DataWeave by adding `output text/plain ---` to your script.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+[NOTE]
+---
+The HTTP RFC header names are case insensitive. The MuleSoft HTTP Connector changes header names to lowercase characters. However, the DataWeave is key case sensitive. Therefore, when creating the Identifier expression, remember to reference headers in lowercase.
```
kbostrom:
```suggestion
The HTTP RFC header names are case-insensitive. Anypoint Connector for MuleSoft HTTP changes header names to lowercase characters. However, the DataWeave key is case-sensitive. Therefore, when creating the Identifier expression, remember to reference headers in lowercase.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+---
+
+=== Configuring Unbound Identifier Sets
+
+Every Identifier result has one algorithm.You must carefully create the DataWeave expression such that it does not return an unbound or a very large co-domain, as it will require hosting the same amount of algorithms in memory (at least a request for every possible identifier has to be made as algorithms are created lazily).
```
kbostrom: missing space; also, consider rewording the end to use active voice
```suggestion
Every Identifier result has one algorithm. You must carefully create a DataWeave expression that does not return an unbound or a very large co-domain, which requires hosting the same number of algorithms in memory (at least a request for every possible identifier has to be made, because algorithms are created lazily). 
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+[NOTE]
+---
+The HTTP RFC header names are case insensitive. The MuleSoft HTTP Connector changes header names to lowercase characters. However, the DataWeave is key case sensitive. Therefore, when creating the Identifier expression, remember to reference headers in lowercase.
+---
+
+=== Configuring Unbound Identifier Sets
+
+Every Identifier result has one algorithm.You must carefully create the DataWeave expression such that it does not return an unbound or a very large co-domain, as it will require hosting the same amount of algorithms in memory (at least a request for every possible identifier has to be made as algorithms are created lazily). 
+
+For example, suppose the DataWeave expression uses the IP as the identifier in a Mule runtime engine that is public to the Internet. If every public IPv4 IP on the internet makes a request to this Mule instance, there will be 3,706,452,992 algorithms running in a single Mule instance.
```
kbostrom: "the IP" what? IP needs an object here.... IP address?

kbostrom: Do we need to say "...a Mule runtime engine that is public to the Internet" or can we use "...a public Mule runtime engine instance"?

kbostrom: GLOBALLY lowercase "internet"
(except when spelling out the name IoT)

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+---
+
+=== Configuring Unbound Identifier Sets
+
+Every Identifier result has one algorithm.You must carefully create the DataWeave expression such that it does not return an unbound or a very large co-domain, as it will require hosting the same amount of algorithms in memory (at least a request for every possible identifier has to be made as algorithms are created lazily). 
+
+For example, suppose the DataWeave expression uses the IP as the identifier in a Mule runtime engine that is public to the Internet. If every public IPv4 IP on the internet makes a request to this Mule instance, there will be 3,706,452,992 algorithms running in a single Mule instance. 
+
+On an average of 250 bytes per algorithm, this approximately amounts to 1 terabyte in Rate Limit algorithms. Therefore, MuleSoft recommends that you use such a DataWeave expression that resolves to a finite number of identifiers, and to keep the resulting set as small as possible.
```
kbostrom:
```suggestion
At an average of 250 bytes per algorithm, this amounts to approximately 1 terabyte of Rate Limit algorithms. Therefore, use a DataWeave expression that resolves to a finite number of identifiers to keep the resulting set as small as possible.  
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+For example, suppose the DataWeave expression uses the IP as the identifier in a Mule runtime engine that is public to the Internet. If every public IPv4 IP on the internet makes a request to this Mule instance, there will be 3,706,452,992 algorithms running in a single Mule instance. 
+
+On an average of 250 bytes per algorithm, this approximately amounts to 1 terabyte in Rate Limit algorithms. Therefore, MuleSoft recommends that you use such a DataWeave expression that resolves to a finite number of identifiers, and to keep the resulting set as small as possible.  
+
+=== Configure Rate Limiting Policy for Clusters
+
+Let’s take the same configuration example of 3 requests and a window that starts exactly at 12:00:00, is reset every 10 seconds, and has a 2-node Mule cluster. Both nodes start and end their windows at the same time, and the cluster allows 3 requests per window in total:
```
kbostrom: Reword to not use "Let's".

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+image::rate-limiting-configure-clusters.png[Rate Limting Policy Configure Clusters,60%,75%]
+
+As the policy is clusterized, the whole cluster accepts 3 requests. If the clusterizable policy is turned off, the Mule cluster can accept 6 requests per window, that is 3 requests per node.
```
kbostrom: GLOBALLY, do not use "as" when you mean "because":
```suggestion
Because the policy is clusterized, the whole cluster accepts three requests. If the clusterizable policy is turned off, the Mule cluster can accept six requests per window: that is, three requests per node.
```
Ugh! that "10 seconds" is messing up the whole spelling/no spelling of amounts and values....

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+Distributed counters impact performance due to the need for synchronization between the nodes. The policy uses caching mechanisms to predict the behavior and maximize throughput. However, in a worst case scenario, you can expect higher latency. Therefore, whether you use clusterizable configuration must depend solely on your use case.
```
kbostrom: Does this mean the same thing?
```suggestion
To avoid distributed counters negatively impacting performance due node synchronization needs, the policy uses caching mechanisms to predict the behavior and maximize throughput. However, some scenario nevertheless result in higher latency, so be careful in your use of clusterizable configuration.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+[NOTE]
+---
+For this feature, the Mule runtime engine instance must be running as part of a Mule cluster.
```
kbostrom: For what feature?
And don't we mean "For you to benefit from <this feature>, you Mule instance must run as part of a cluster"?

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+---
+
+== Configuring the Policy
+
+When you configure your Rate Limiting policy, you must consider certain aspects of your environment to help you derive the most value from the policy, as explained in this section.
```
kbostrom: Per style ref, omit all self-referential wordings to this section, this topic, this book....
```suggestion
When you configure your Rate Limiting policy, you must consider certain aspects of your environment to help you derive the most value from the policy.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+As the policy is clusterized, the whole cluster accepts 3 requests. If the clusterizable policy is turned off, the Mule cluster can accept 6 requests per window, that is 3 requests per node.
+
+Distributed counters impact performance due to the need for synchronization between the nodes. The policy uses caching mechanisms to predict the behavior and maximize throughput. However, in a worst case scenario, you can expect higher latency. Therefore, whether you use clusterizable configuration must depend solely on your use case.
+
+[NOTE]
+---
+For this feature, the Mule runtime engine instance must be running as part of a Mule cluster.
+---
+
+== Configuring the Policy
```
kbostrom: I know you have a template, but a more precise title would be....
```suggestion
== Configuring Your Rate Limiting Policy
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+When you configure your Rate Limiting policy, you must consider certain aspects of your environment to help you derive the most value from the policy, as explained in this section.
+
+=== Cluster or Standalone
```
kbostrom: Similarly:
```suggestion
=== Choosing Between a Cluster or Standalone Configuration
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+You might have decentralized processing in your environment, with the following setup:
+
```
kbostrom, line 131: The following list has two fragments and two sentences; please try to make them all the same/parallel.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* N servers for the same API
```
kbostrom: "N servers" means what? Are you using N as the variable _n_?
But if it is a variable, how is it a defined setup?

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* Each server has its own backend
```
kbostrom:
```suggestion
* Each server has its own backend.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* The amount of requests that can be served is limited only by the backend
```
kbostrom:
```suggestion
* The number of requests that can be served is limited only by the backend.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* Fastest response time
+
+In such a scenario, you do not need to run the policy in a clustered setup. Simply set the policy limits of the policy lower than the backend capacity of the weakest of the nodes. Additionally, a load balancer might be useful in case a node goes down.	
+
+On the other hand, if you have centralized processing in your environment, with the following setup:
```
kbostrom: ditto re. replacing "on the other hand"

kbostrom: This intro sentence is incomplete when using "if". Reword something like this:
```suggestion
Alternatively, you might have centralized processing in your environment, with the following setup:
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+* N servers for the same API
```
kbostrom: ditto re "N"

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+* Single backend to which all of the proxies connect
+* Possible load balancer in front of the proxies
+
+In such a scenario, you do not need a cluster. If q is the maximum capacity of the backend, then configure the policy as follows:  where  is a small number, lower than the backend capacity.
```
kbostrom: Wait...this is supposed to be an opposite example to the first, and yet this one likewise does not require a cluster.
Reread these two scenario paragraphs and clarify whether you are comparing or just describing similar setups.

kbostrom: What is "q" and reword sentence to use proper grammar:

If <this>, then configure <that> to have a value lower than the backend capacity.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+If there are no load balancers in your environment, the cluster mode is recommended over a standalone, because you cannot configure beforehand how much traffic each node will handle. The Rate Limiting policy is designed to work both on a perfectly balanced workload, or on a completely uneven one. The backend does not receive any extra requests and thus does not exceed the maximum capacity that it can handle.
```
kbostrom: Many issues here:

- Avoid using "there is" and "there are" as implied sentence subjects. See style ref.
- Passive voice
- Lack of parallel wording
- Verbosity


```suggestion
If your environment does not include load balancers, use a cluster instead of a standalone instance to be certain that your nodes can manage different levels of traffic. The Rate Limiting policy is designed to work under both balanced and unbalanced workloads. Because the backend does not receive any extra requests, its maximum capacity is not exceeded.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+=== Window Sizes in Clusters
```
kbostrom: similarly
```suggestion
=== Choosing Window Sizes for Cluster Nodes
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+In clusters, the nodes must share information for consistency across the cluster. The sharing process adds latency that must be taken into account when reviewing performance.
```
kbostrom:
```suggestion
In clusters, the nodes must share information to achieve consistency across the cluster. The sharing process adds latency that must be considered when reviewing performance.
```
"to take into account" is an idiom
Note that this is still in passive voice and should be rewritten to be active: GLOBAL

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+In a worst case scenario, the number of penalized requests with latency due to cluster consistency is constant and independent from the actual size of the configured quota. Consequently, the smaller the window, the greater the percentage of potentially delayed requests.
```
kbostrom: The worst-case scenario is the constancy and independence? Clarify this opening sentence: what causes the worst case, and what is the negative effect experienced by the user?

kbostrom: Seems to me you could simplify the previous para into a statement that the readers should choose windows large enough to avoid request latency; in fact, simpler still (reading ahead), this whole topic can be summed up as follows:

Configure your environment to use processing windows longer than one minute, to prevent the latency potential caused by information sharing among nodes in a cluster.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+=== Window Sizes in Clusters
+
+In clusters, the nodes must share information for consistency across the cluster. The sharing process adds latency that must be taken into account when reviewing performance.
+
+In a worst case scenario, the number of penalized requests with latency due to cluster consistency is constant and independent from the actual size of the configured quota. Consequently, the smaller the window, the greater the percentage of potentially delayed requests. 
+
+Therefore, MuleSoft recommends that you set up only window sizes greater than one minute in Rate Limiting and Rate Limiting SLA policy configurations for the cluster scenario.
+
+=== Persistence
```
kbostrom: Look at all of the topic titles at the same level and be sure that they are truly peer subtopics and that they are worded in a parallel manner.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+You can configure Rate Limiting policy to use big windows sizes, such asdays, months, and years. For example, suppose as a client you want to allow your user ‘X’ to consume 1 million requests per year. You cannot predict whether the node will be up the entire period or will need maintenance, which may result in restarting the Mule runtime engine.
```
kbostrom:
```suggestion
You can configure your Rate Limiting policy to use windows that persist as long as days, months, and years. For example, suppose you want to allow your user to consume 1 million requests per year, but you cannot ensure that the node will be up the entire period or that the Mule runtime engine won't be restarted. 
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+You can configure Rate Limiting policy to use big windows sizes, such asdays, months, and years. For example, suppose as a client you want to allow your user ‘X’ to consume 1 million requests per year. You cannot predict whether the node will be up the entire period or will need maintenance, which may result in restarting the Mule runtime engine. 
+
+The algorithm has been running for several months, so the client will lose critical information. Persistence solves this problem by periodically saving the current policy state. In case of a redeployment or a restart, the algorithms are recreated from the last known persisted state or started from a clean state.
+
+Although persistence is enabled by default, you can turn it off by setting the following property to false:
+
+`throttling.persistence_enabled`
+
+You can also tweak the persistence frequency rate, which has a default of 10 seconds:
```
kbostrom, line 166:
```suggestion
You can also tweak the persistence frequency rate, which has a default of 10 seconds:
+
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+---
+
+== Configuring Policy Parameters
+
+When you apply the Rate Limiting policy to your API from the UI, you can configure the following parameters:
+
+[%header%autowidth.spread,cols="a,a,a"]
+|===
+|Parameter | Description | Example
+|Identifier | Specifies a selector key using a Dataweave or regular string | `#[attributes.method]` This creates one group for each available method in HTTP, and for example, the policy will rate limit GET requests independently from POST requests.
```
kbostrom:
```suggestion
|Identifier | Specifies a selector key using a Dataweave or regular string. | `#[attributes.method]` creates one group for each available method in HTTP; for example, the policy rate-limits GET requests independently from POST requests.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+|# of Reqs | The quota available per window | A positive number, for example: 3.
```
kbostrom: When I look at the print view for this, # of Reqs is malformed (not formatted/coded correctly?) 
Please check.

kbostrom: Note that the wordings for Description are not parallel: some are just a noun ("The quota...") and others begin with verbs ("Specifies a selector key...").
Try to reword to make them parallel. All peer texts should be parallel.

kbostrom: "A positive number, for example: 3."     >
    A positive number, for example, 3
or
   A positive number like 3
or
   A positive number

...because everyone knows what a positive number is. And perhaps you need "positive whole number"?

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+|Time Period | The amount of time for which the quota is to be applied | A positive number, for example: 10
```
kbostrom: ditto for example 10
Won't it confuse readers that you choose first 3 and then 10 for examples of positive numbers?

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+|Time Unit | The time unit for the Time Period in milliseconds, seconds, minutes, or hours |Minutes
```
kbostrom:
```suggestion
|Time Unit | Milliseconds, seconds, minutes, or hours |Minutes
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+|Clusterizable | Shares the quota defined for a group among all of the nodes of a Mule cluster | checked/unchecked
```
kbostrom:
```suggestion
|Clusterizable | Shares the quota defined for a group among all of the nodes of a Mule cluster. | checked or unchecked
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+| Expose Headers| Defines whether to expose the  <<x-ratelimit headers,response-header>> as part of the response | checked/unchecked
```
kbostrom: Is this formatted correctly?

<<x-ratelimit headers,response-header>>

kbostrom: checked or unchecked

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+|===
+
```
kbostrom, line 190: Hmph...we are trying to avoid including FAQs in our docs.
Talk with your product lines leads peers about this.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+== FAQ
+
+*When does the window start?*
+The window starts on the first request after the policy has been successfully applied.
```
kbostrom:
```suggestion
The window starts with the first request after the policy is successfully applied.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+*What type of window does the algorithm use?*
+It uses a fixed window.
+
+*What happens when the quota is exhausted?*
+The algorithm is created on demand, when the first request is received. This event fixes the time window. Each request consumes quota from the current window until the time expires.
```
kbostrom: "consumes quota"? How does something consume a quota? It can consume something until it reaches the limits of a quota....

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+*When does the window start?*
+The window starts on the first request after the policy has been successfully applied.
+
+*What type of window does the algorithm use?*
+It uses a fixed window.
+
+*What happens when the quota is exhausted?*
+The algorithm is created on demand, when the first request is received. This event fixes the time window. Each request consumes quota from the current window until the time expires. 
+
+When the quota is exhausted, the Rate Limiting policy rejects the request. When the time window closes, the quota is reset and a new window of the same fixed size starts.
```
kbostrom: What actually stops processing: the quota being exhausted, the rejection, or the time expiring?

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+*What happens if I define multiple limits?*
+The policy creates one algorithm for each limit with its quota per time window configuration. Therefore, when multiple limits are configured, every algorithm must have available quota in its current window for the request to be accepted.
```
kbostrom: Please reread the first sentence. I can't even begin to figure out how to reword this for clarity. Ask a peer writer for help, perhaps.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+[[response-header]]
+*What does each response header mean?*
+Each response header has information about the current state of the request:
+
+* X-Ratelimit-Remaining: The amount of available quota
+* X-Ratelimit-Limit: The maximum available requests per window
+* X-Ratelimit-Reset: The remaining time, in milliseconds, until a new window starts
+
+MuleSoft advises that you expose these headers only if the API is used within the organization. In a public API scenario, this data can be used maliciously to timely burst in traffic that leaves your API without any quota left.
```
kbostrom:
```suggestion
MuleSoft advises that you expose these headers only if the API is used within the organization. In a public API scenario, this data can be used maliciously to burst in traffic that leaves your API without any quota left.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+*Can I configure a Mule Cluster in CloudHub?*
```
kbostrom:
```suggestion
*Can I configure a Mule cluster in CloudHub?*
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+No, the feature is available only for RTF, Hybrid and Standalone Mule setups.
```
kbostrom: "the feature" = a cluster?

kbostrom:
```suggestion
No, the feature is available only for RTF, hybrid, and standalone Mule setups.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+*When should I use Rate Limiting instead of Rate Limiting SLA or Spike Control?*
+Rate Limiting and Rate Limiting SLA policies must be used for accountability and to enforce a hard limit to a group (using the Identifier in Rate Limiting) or to a Client application (using Rate Limiting SLA). If you want to protect your backend, use the Spike Control policy instead.
```
kbostrom:
```suggestion
Use Rate Limiting and Rate Limiting SLA policies for accountability and to enforce a hard limit to a group (using the Identifier in Rate Limiting) or to a client application (using Rate Limiting SLA). If you want to protect your backend, use the Spike Control policy instead.
```

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+
+== See Also
+
+* xref:tutorial-manage-an-api.adoc[Applying a Policy and SLA Tier]
+* xref:delete-sla-tier-task.adoc[Removing SLA Tiers on API Manager]
```
kbostrom: the tiers are ON API Manager, or you are removing them FROM?

kbostrom (dismissed):
Second part of CU edit, per scope of request

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: "Clusterized" is the name of the configuration in the UI. Made the other change. Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: The same example applies to each of the scenarios. The heading level is correct of each of the sub-topic.

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: It's commented out. Not needed for now.

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: In clusters, the nodes share information to achieve consistency across the cluster. Therefore, when reviewing performance, consider the latency added on by the information sharing.

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,237 @@
...
+The Rate Limiting quota applies to every request per bucket or group.
+* Configured for an obligatory HTTP header
++
+Each header has its own quota. Header values are case-sensitive. Quotas are created using the _lazy creation_ strategy.
+* Configured for a non-obligatory HTTP header
++
+Custom header, payload, query parameter, or expression values each have their own quotas.  
++
+The `identifier`, if not sent in the request, defaults to an empty value, having its own quota. This behavior allows the Rate Limiting policy to be applied to an API consumed by uncontrolled clients, and at the same time accommodates special buckets for the clients sending the `identifier` (which might be internal controlled services).
+//ask Mariano what does the braces refer to? What might be internal controlled services
```
mtqp: I try to refer to the fact that you could say "ok I've created a special bucket just foryou, please you identifier {yourName}, for the rest, they shall default to the 'uncontrolled clients' bucket". The client with the specific identifier be a 'controlled' one.
I think you can remove the comment within braces, it use to make sense in a previous iteration. I

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,237 @@
...
+The following example shows the order of events that occur over a period of time using the `identifier` `#[attributes.method]` for a limit of 3 requests every 10 seconds:
+
+image::rate-limiting-configure-identifiers-example.png[Rate Limiting Configure Identifiers Example,70%,85%]
+In the example:
+
+* Every HTTP method is allowed 3 requests every 10 seconds (in this example, only GET and POST requests are made to the API).
+* The Rate Limiting policy works in a fixed-window fashion.
++
+For more information, see the window size bracket.
+//Mariano--See what window size bracket, and how does this explain "fixed-window fashion"?
```
mtqp: You need to look at the picture to see the bracket. The "fixed-window fashion" refers to that bracket, one that starts and closes at well defined points in time.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+*When does the window start?*
+The window starts with the first request after the policy is successfully applied.
+
+*What type of window does the algorithm use?*
+It uses a fixed window.
+
+*What happens when the quota is exhausted?*
+The algorithm is created on demand, when the first request is received. This event fixes the time window. Each request consumes the request quota from the current window until the time expires. 
+//Mariano--what quota--request right?
```
mtqp: Yes

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+You can configure the policy for multiple groups of requests by using identifiers in the policy configuration. Each group has a separate available quota for its window. 
+
+To understand how the Rate Limiting policy works, consider an example in which the configuration of 3 requests every 10 seconds allows or restricts incoming requests, based on the quota available in that window:
+
+image::rate-limiting-generic-example.png[Rate Limiting Policy Example,70%,85%]
+
+In the first window, because the quota is reached with the third request, all subsequent requests are rejected until the window closes. In the second window, only two of the three requests are processed and the quota remaining is dropped after the window time has elapsed.
+
+An accepted request passes through the API to the backend. Conversely, a rejected request displays 1429 status for HTTP (or either 400 or 500 if the API is WSDL) and does not reach the backend.
```
mtqp: '1429' is not a valid status code, the '1' seems to be a type, status code should be 429.

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+A rejected request on the other hand, displays a “429 status for HTTP, ” (or either 400 or 500 if the API is WSDL) and does not reach the backend.
```
mtqp, line 40: I missed this typo in the gdoc, sorry! Could you change it for **“429 status for HTTP”,**

modules/ROOT/pages/rate-limiting.adoc
```diff
@@ -0,0 +1,237 @@
...
+In such a scenario, you do not need to run the policy in a clustered setup. Simply set the policy limits of the policy lower than the backend capacity of the weakest of the nodes. Additionally, a load balancer might be useful in case a node goes down.	
+
+Alternatively, you might have centralized processing in your environment, with the following setup:
+
+* You have more than one servers for the same API.
+* You have a single backend to which all of the proxies connect.
+* You have a load balancer in front of the proxies
+
+In such a scenario, you do not need a cluster. However, you must then configure the policy to have a value lower than the maximum capacity of the backend.
+//Mariano--reworded this sentence. Please check.
```
mtqp: Looks good!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/rate-limiting.adoc
priyaayyar: Done! Thanks!

kbostrom (approved):
Unedited reapproval

mtqp (approved):

PR #307, opened 2020-11-16
DOCS-5034 added about export API

priyaayyar:


modules/ROOT/pages/import-api-task.adoc
```diff
@@ -3,7 +3,9 @@ ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-You can import a previously exported API to share it across your organization or to use it at a later time. You can import an API's ZIP file from your local file system or server.
+You can import a previously exported API to share it across your organization or to use it at a later time. You can import an API's ZIP file from your local file system or server. 
+
+To import an API, you must have a copy of the API ZIP file exported from API Manager.
```
evangelina, line 10: Maybe we can add it as a note or something that brings more attention to this line? The comment is because this is a common root cause for support escalations and it's repeating what it says in the previous sentence in a way. What do you think?

priyaayyar: Done! Thanks!

evangelina (approved):
Approved, thanks for taking the feedback.

kevintroller (approved):
LGTM

PR #310, opened 2020-12-01
DOCS-3314 updated the note

priyaayyar:
with business group requirement

### DIFF
--- a/modules/ROOT/pages/import-api-task.adoc
+++ b/modules/ROOT/pages/import-api-task.adoc
@@ -7,7 +7,7 @@ You can import a previously exported API to share it across your organization or
 
 [NOTE]
 
-For the import to work successfully, you must copy of the API ZIP file exported from API Manager. You must import the API to the same business group as the one from which you exported in the source location.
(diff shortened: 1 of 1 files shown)


sbehrends (dismissed):

modules/ROOT/pages/import-api-task.adoc
```diff
@@ -7,7 +7,7 @@ You can import a previously exported API to share it across your organization or
 
 [NOTE]
 
-To import an API, you must have a copy of the API ZIP file exported from API Manager.
+For the import to work successfully, you must copy of the API ZIP file exported from API Manager. You must import the API to the same business group as the one from which you exported in the source location.
```
Cristian-Venticinque: you must copy **of** the > you must copy the ?

I'd suggest reworking the first sentence to something like: 
You can only import API ZIP files that were exported from API Manager.

Cristian-Venticinque (dismissed):
Please address comments before merging

modules/ROOT/pages/import-api-task.adoc
priyaayyar: Made the first change but left the rest as is. Thanks!

Cristian-Venticinque (approved):

PR #319, opened 2020-12-18
DOCS-9653 Update API Manager MiDPs docs

chris-campbell9:
JIRA:
https://www.mulesoft.org/jira/browse/DOCS-9653

Note:
The JIRA also includes a request for graphics edits. I'll add these edits to a new PR after I receive my OmniGraffle license (requested on 12/10).


modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
 
 *Best Practice*
 
-To secure your APIs by leveraging the multiple client providers feature, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. 
+To secure your APIs, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. However, check the possible impact on managed APIs and their consumers. 
 
-However, before changing any existing client providers, ensure that you check the impact on managed APIs and their consumers. Before you implement multiple client providers, see the <<guidelines,guidelines>>.
+Avoid using the same IdP in different environments.
```
franciscoSoler: This message is not completely accurate, the best practice is to use different IDPs for productive and non-productive environments. 
It is okay to associate the same IDP or with several productive environments or with several non productive ones. We need to avoid the association of a certain IDP with productive and non-productive envs.

franciscoSoler (commented):
Nice work, just a small comment

modules/ROOT/pages/configure-multiple-credential-providers.adoc
chris-campbell9: Hi @franciscoSoler, thanks! I think I fixed it. But please let me know if I'm still off.

franciscoSoler (dismissed):
LGTM

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
 
 *Best Practice*
 
-To secure your APIs by leveraging the multiple client providers feature, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. 
+To secure your APIs, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. However, check the possible impact on managed APIs and their consumers. 
 
-However, before changing any existing client providers, ensure that you check the impact on managed APIs and their consumers. Before you implement multiple client providers, see the <<guidelines,guidelines>>.
+Avoid using the same IdP in production and non-production environments. However, the same IdP can be used in multiple production environments, or in multiple non-production environments.
```
priyaayyar:
```suggestion
Avoid using the same IdP in production and non-production environments. You can use the same IdP in multiple production environments, or in multiple non-production environments.
```

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -99,9 +100,9 @@ Returning to the Good Weather use case, imagine that the company uses the follow
 
 image::multiple-IdP-user-flow.png[User Flow for Configuring Multiple External Client Providers]
 
-As shown in the diagram, you first configure your external client provider and assign it to the appropriate business groups and environments. When you create or modify an API within a specific business group or environment, the new client provider is displayed. You can reselect the appropriate client provider. 
+As shown in the diagram, you first configure your external client provider and assign it to the appropriate environments. When you create or modify an API within a specific environment, the new client provider is displayed. You can reselect the appropriate client provider. 
 
-When you promote an API to another environment, the target environment must have the client provide configured as in the source environment. Otherwise, the promotion does not complete successfully. 
+When you promote an API to another environment, the target environment must have the client provider configured as in the source environment. Otherwise, a promoted API's IdP policy may not match the API's associated IdP. To resolve this mismatch, delete the IDP policy.
```
priyaayyar:
```suggestion
When you promote an API to another environment, the target environment must have the client provider configured as in the source environment. Otherwise, a promoted API's IdP policy might not match the API's associated IdP. To resolve this mismatch, delete the IDP policy. 
```

priyaayyar (dismissed):

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -8,17 +8,16 @@ You can use multiple client providers, to help you enforce security and regulati
```
kbostrom, line 8: Hello Kim! When you have a moment, would you take a look at the following PRs:
https://github.com/mulesoft/docs-api-manager/pull/319
https://github.com/mulesoft/docs-access-management/pull/109
Type: CE
Due: Per your convenience

kbostrom: FYI, I am typically asked to review the **green** (changed) text, but sometimes **all** text (green and white).

The blue text is not expanded because it has not be touched in a long while. However, it is likely to need editing as well...but only after the writer tries to clean it up first. Which turns it green...lol.

Lemme' know if this is confusing and I will try to reexplain.

kbostrom: So...Imma' edit the green here. If you want the white edited, send it on back; since you asked for a CE and not a CU+CE, I am guessing you are on a tight schedule.

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -8,17 +8,16 @@ You can use multiple client providers, to help you enforce security and regulati
 
 You can use an Anypoint Platform native client provider (default) or configure an external client provider. To assign separate client providers for different organizations and environments, you must first enable client applications to be authorized using OAuth for the client providers that you want to implement in your organization.
 
-In API Manager 2.2.13 and earlier, only a single IdP was supported: either an external IdP or the default IdP (Anypoint Platform native client provider). With the addition of the multiple client providers feature, API Manager 2.2.14 and later supports either one default IdP or one or more external IdPs.
+In API Manager 2.2.13 and earlier, only a single identity provider (IdP) was supported: either an external IdP or the default IdP (Anypoint Platform native client provider). With the addition of the multiple client providers feature, API Manager 2.2.14 and later supports either one default IdP or one or more external IdPs.
```
kbostrom: We try to stick with present tense as much as possible, per Style Ref:
```suggestion
In API Manager 2.2.13 and earlier, only a single identity provider (IdP) is supported: either an external IdP or the default IdP (Anypoint Platform native client provider). With the addition of the multiple client providers feature, API Manager 2.2.14 and later supports either one default IdP or one or more external IdPs.
```
Part of what I do (arguably the most important part) is consider how something might be more simply said. You are always free to reject these suggestions, and you should always be sure that they don't inadvertently change your intended meaning or impair technical accuracy. I am not as product-savvy as your audience might be, but I do focus on the language and how it strikes me as a First Reader.

SO, all that said, your text is FINE as-is, but consider this:
```suggestion
API Manager 2.2.14 introduces multiple client identity provider (IdP) support, enabling your environment to use either the default Anypoint Platform native client IdP or one or more external client IdPs.
```

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -8,17 +8,16 @@ You can use multiple client providers, to help you enforce security and regulati
...
 
 You use multiple client providers for:
 
-* Different business groups
 * Different environments, such as production, QA, and sandbox
 * Internal and external APIs
 
 == How This Works
```
kbostrom: If I were skimming the TOC, I might balk at this subtitle, not being too certain to what "This" refers. We can no longer assume that our audiences actually read chronologically....

Always attempt to make topic titles precisely and concisely represent the actual content within, relying on the text immediately following to expand on things like abbreviations or conditions or caveats to the title if necessary.

Consideration:
```suggestion
== How Using Multiple External Client Providers Works
```

kbostrom: Consider also that you might not need this subtopic as separate from the intro to the topic set.

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
 
 *Best Practice*
 
-To secure your APIs by leveraging the multiple client providers feature, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. 
+To secure your APIs, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. However, check the possible impact on managed APIs and their consumers.
```
kbostrom: Question about "...and add the new providers to other environments, such as sandbox**, and QA.**"

Are we suggesting adding the providers to both QA and other environments, or are we calling QA another environment? Maybe put QA first?:

```suggestion
To secure your APIs, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to QA and to other environments, such as the sandbox. However, check the possible impact on managed APIs and their consumers. 
```

kbostrom: I also wonder about the final sentence, which seems here like an afterthought and is not very specific about how to check this impact or what the impact might be.

Should this be more of a prerequisite to assigning the existing client provider elsewhere? Should we include more details about exactly what we are warning the audience against doing?

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
 
 *Best Practice*
 
-To secure your APIs by leveraging the multiple client providers feature, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. 
+To secure your APIs, create one external client provider per environment. Assign the existing client provider to your production environment and add the new providers to other environments, such as sandbox, and QA. However, check the possible impact on managed APIs and their consumers. 
 
-However, before changing any existing client providers, ensure that you check the impact on managed APIs and their consumers. Before you implement multiple client providers, see the <<guidelines,guidelines>>.
+Avoid using the same IdP in production and non-production environments. You can use the same IdP in multiple production environments, or in multiple non-production environments.
```
kbostrom:
```suggestion
Avoid using the same IdP in production and nonproduction environments. You can use the same IdP in multiple production environments or in multiple nonproduction environments.
```

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
...
 
+Before implementing multiple client providers, see the <<guidelines,guidelines>>.
```
kbostrom, line 55: Again, this seems to be an afterthought when it sounds like more of a prerequisite.

[I am remembering now that I have commented about these things with the previous writer of this text; hoping that we can make some structural changes along the way....]

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
...
 
 [[guidelines]]
 == Guidelines for Implementing Multiple Client Providers
```
kbostrom, line 58: Per previous comments, I suggest that we look further into the order and titling of subtopics here.
I would call these requirements: indeed, prerequisites to implementing multiple providers and something that should precede any instructions or be mentioned at the beginning of any conceptual descriptions.

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,19 +48,21 @@ If you have external client providers such as PingFederate configured for your e
...
 
 Before you implement multiple client providers in Anypoint Platform, carefully review and understand the following points:
 
-* When you change the client provider for an existing API instance, ensure that the API:
-** Has no contracts
-** Has no client provider-related policies applied, such as PingFederate and OpenID policies
+* You cannot change a client provider's existing API instance when that API:
+** Has contracts.
+** Has client provider-related policies applied, such as PingFederate and OpenID policies.
+** Belongs to an API group.
```
kbostrom: No periods, per Style Ref.

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -71,7 +72,7 @@ You can then reassign the appropriate external client provider for that API.
 ** New APIs use the new external IdP.
 * After you assign an external client provider to an API, you cannot return to using the default native Anypoint Platform client provider.
 +
-You can user either the default IdP or one or more external IdPs. You cannot use the default provider and an external client provider together.
+You can use either the default IdP or one or more external IdPs. You cannot use the default provider and an external client provider together. To change an API's IdP from an external provider to an Anypoint Platform client provider, remove every IdP from the environment in which the API belongs.
```
kbostrom: I suggest that you GLOBALLY search the material in these topics and choose to refer to one but not both of the following after initially introducing the two terminologies:

the default IdP or one or more external IdPs
the default native client [identity] provider or one or more external client [identity] providers

In one of my previous editing lives, we recommended that once you provide an abbreviation, and it is more commonly used than the full spelling, then you should use only it in following text (in that topic, anyway). So maybe you want to use just "default IdP" and "external IdP"?

kbostrom: Based on the first sentence, I would omit the second: "either" should make clear that you cannot use both.

kbostrom: Clarify how the third sentence relates to the previous two?
Also...WHY or under what circumstances would the audience want or need to change the IdP?

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -89,7 +90,7 @@ Based on your role and the needs of your organization, you might perform the fol
...
 | Review to which client provider an application belongs on the client administrator page to be able to differentiate between different applications with the same name. | API Manager Administrator| Access Management | None
 | Configure an API in API Manager to automatically map to a default IdP configured by the business group administrator. The owner can also select a different IdP, if available. | API Owner | API Manager | None
-| When promoting or importing an API between environments, specify an IdP from the target environment. | API Owner | API Manager | If there is just one (default) IdP specified for the target environment, it is automatically assigned to the API on promotion. 
+| When promoting or importing an API between environments, specify an IdP from the target environment. | API Owner | API Manager | If there is just one (default) IdP specified for the target environment, it is automatically assigned to the API.
```
kbostrom: "Promoting" and "importing" don't really work with "between". Do we really mean "from one environment to another"? Also... "promoting"?

kbostrom: API owner, API manager
lowercase

kbostrom: See if you can omit the "Notes" column, which is a vague distinction.
In the case of this table row, the Note could be part of the task statement: When importing, if the target has only one, it is assigned to the API; otherwise, specify which target IdP to use.

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -99,9 +100,9 @@ Returning to the Good Weather use case, imagine that the company uses the follow
 
 image::multiple-IdP-user-flow.png[User Flow for Configuring Multiple External Client Providers]
 
-As shown in the diagram, you first configure your external client provider and assign it to the appropriate business groups and environments. When you create or modify an API within a specific business group or environment, the new client provider is displayed. You can reselect the appropriate client provider. 
+As shown in the diagram, you first configure your external client provider and assign it to the appropriate environments. When you create or modify an API within a specific environment, the new client provider is displayed. You can reselect the appropriate client provider.
```
kbostrom, line 103: Clarify or simplify: you first configure and assign, but if you add or modify an API in an environment, you [might?] need to change to a new provider?

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -99,9 +100,9 @@ Returning to the Good Weather use case, imagine that the company uses the follow
 
 image::multiple-IdP-user-flow.png[User Flow for Configuring Multiple External Client Providers]
 
-As shown in the diagram, you first configure your external client provider and assign it to the appropriate business groups and environments. When you create or modify an API within a specific business group or environment, the new client provider is displayed. You can reselect the appropriate client provider. 
+As shown in the diagram, you first configure your external client provider and assign it to the appropriate environments. When you create or modify an API within a specific environment, the new client provider is displayed. You can reselect the appropriate client provider. 
 
-When you promote an API to another environment, the target environment must have the client provide configured as in the source environment. Otherwise, the promotion does not complete successfully. 
+When you promote an API to another environment, the target environment must have the client provider configured as in the source environment. Otherwise, a promoted API's IdP policy might not match the API's associated IdP. To resolve this mismatch, delete the IDP policy.
```
kbostrom: Promote?

kbostrom:
```suggestion
To promote an API from one environment to another, either both environments must use the same client provider or you must delete [which?] IdP policy to prevent a policy mismatch.
```

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -112,12 +113,12 @@ To configure multiple external client providers:
 . From *Access Management*, first enable xref:access-management::managing-api-clients.adoc[client management].
 +
 As an Anypoint Administrator, you must first enable client applications to be authorized using OAuth for the client providers to be implemented in the organization.
-. From *Access Management*, xref:access-management::environments.adoc#add-an-idp-to-an-existing-environment[assign client providers] to environments and business groups.
+. From *Access Management*, xref:access-management::environments.adoc#add-an-idp-to-an-existing-environment[assign client providers] to environments.
 +
-As an Anypoint Administrator, you configure a client provider and assign it to specific business organizations and environments.
+As an Anypoint Administrator, you configure a client provider and assign it to specific environments.
```
kbostrom:
```suggestion
As an Anypoint Administrator, configure a client provider and assign it to specific environments.
```

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -112,12 +113,12 @@ To configure multiple external client providers:
...
 . From *API Manager*, xref::create-instance-task.adoc[create or modify] APIs using the new IdPs.
 +
-As an API Owner create an API, import an API from Anypoint Exchange or modifie an API in API Manager. Optionally, you can promote an API from one environment to another, and import or export an API, as described in the user flow diagram in the previous section. 
+As an API Owner create an API, import an API from Anypoint Exchange or modify an API in API Manager. Optionally, you can promote an API from one environment to another, and import or export an API, as described in the user flow diagram in the previous section.
```
kbostrom:
```suggestion
As an API owner, create an API, import an API from Anypoint Exchange, or modify an API in API Manager. Optionally, either promote an API from one environment to another or import or export an API as described earlier.
``

**Review by kbostrom — DISMISSED** [review]
CE with a little CU thrown in for good measure

File: modules/ROOT/pages/configure-multiple-credential-providers.adoc (hunk 15 above)
Comment (franciscoSoler) [reply modules/ROOT/pages/configure-multiple-credential-providers.adoc]: Just an example trying to clarify the `[which?]` part of your suggestion:

When promoting an api instance with, let say, a pingFederate IDP and the `pingFederate access token enforcement` policy (IDP policies match with the IDP type) to an environment with openAM IDPs, the mentioned policy has to be removed for the promoted api to work because the promoted api will be associated with an openAM IDP.  Later, the user may want to apply the policy `openAM access token enforcement` to protect that new api.

If the target environment were associated with any pingFederate IDP and if the customer chose any of those IDPs to be associated with the promoted API, there wouldn't be any conflict with any applied policy because there wasn't any IDP type change.

Comment (chris-campbell9) [reply modules/ROOT/pages/configure-multiple-credential-providers.adoc]: Fixed! Thanks.

File: modules/ROOT/pages/configure-multiple-credential-providers.adoc (hunk 9 above)
Comment (chris-campbell9) [reply modules/ROOT/pages/configure-multiple-credential-providers.adoc:55]: Thanks, Kim. Good point! I'm thinking that the Guidelines section doesn't really list prerequisites, but rather it lists best practices.

File: modules/ROOT/pages/configure-multiple-credential-providers.adoc (hunk 10 above)
Comment (chris-campbell9) [reply modules/ROOT/pages/configure-multiple-credential-providers.adoc:58]: Thanks, Kim. Good point! I'm thinking that the Guidelines section doesn't really list prerequisites, but rather it lists best practices.

File: modules/ROOT/pages/configure-multiple-credential-providers.adoc (hunk 13 above)
Comment (chris-campbell9) [reply modules/ROOT/pages/configure-multiple-credential-providers.adoc]: Hi Kim, the "Notes" column mostly functions as such. I think the example you describe is good, but the rest of the column is generally more notes-like.

The other two suggestions are fixed. Thanks!

File: modules/ROOT/pages/configure-multiple-credential-providers.adoc [hunk 18]
```diff
@@ -112,12 +113,13 @@ To configure multiple external client providers:
...
-. From *Access Management*, xref:access-management::environments.adoc#add-an-idp-to-an-existing-environment[assign client providers] to environments and business groups.
+. From *Access Management*, xref:access-management::environments.adoc#add-an-idp-to-an-existing-environment[assign client providers] to environments.
 +
-As an Anypoint Administrator, you configure a client provider and assign it to specific business organizations and environments.
+As an Anypoint Administrator, configure a client provider and assign it to specific environments.
 . From *API Manager*, xref::create-instance-task.adoc[create or modify] APIs using the new IdPs.
 +
-As an API Owner create an API, import an API from Anypoint Exchange or modifie an API in API Manager. Optionally, you can promote an API from one environment to another, and import or export an API, as described in the user flow diagram in the previous section. 
+As an API owner, create an API, import an API from Anypoint Exchange, or modify an API in API Manager. Optionally, either promote an API from one environment to another or import or export an API as described earlier.
+``
```
franciscoSoler: shouldn't these backquotes be removed?

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -99,9 +100,9 @@ Returning to the Good Weather use case, imagine that the company uses the follow
 
 image::multiple-IdP-user-flow.png[User Flow for Configuring Multiple External Client Providers]
 
-As shown in the diagram, you first configure your external client provider and assign it to the appropriate business groups and environments. When you create or modify an API within a specific business group or environment, the new client provider is displayed. You can reselect the appropriate client provider. 
+As shown in the diagram, you first configure your external client provider and assign it to the appropriate environments. When you create or modify an API within a specific environment, the new client provider is displayed. You can reselect the appropriate client provider. 
 
-When you promote an API to another environment, the target environment must have the client provide configured as in the source environment. Otherwise, the promotion does not complete successfully. 
+To promote an API from one environment to another, either both environments must use the same client provider or you must delete the target environment's IdP policy to prevent a policy mismatch.
```
franciscoSoler: Maybe:
```suggestion
To promote an API from one environment to another, either both environments must use the same client provider type, e.g. PingFederate or OpenAm, or you must delete every client provider-related policies applied to the promoted API to prevent a policy mismatch. 
```

We need to differentiate a client provider instance from their type. A customer can have two pingFederate client providers instances named `PingForDev` and `PingForProd` and one OpenAm instance named `productiveOpenAm`
A user can change the client provider of an Api with PingFederate policies only if the type remains unchanged. In this example the IDP of the previous API can be swapped between `PingForDev` and `PingForProd` without policy mismatches.
If the API were to be associated with `productiveOpenAm` as a result of a promotion and because of having PingFederate-related policies, the API would have a policy mismatch.

modules/ROOT/pages/configure-multiple-credential-providers.adoc
```diff
@@ -49,29 +48,31 @@ If you have external client providers such as PingFederate configured for your e
...
 == Guidelines for Implementing Multiple Client Providers
 
 Before you implement multiple client providers in Anypoint Platform, carefully review and understand the following points:
 
-* When you change the client provider for an existing API instance, ensure that the API:
-** Has no contracts
-** Has no client provider-related policies applied, such as PingFederate and OpenID policies
+* You cannot change a client provider's existing API instance when that API:
+** Has contracts
+** Has client provider-related policies applied, such as PingFederate and OpenID policies
```
franciscoSoler: This particular bullet has a small exception. A user can change a client provider's existing API instance when the provider type doesn't change.
A user can change the IDP of an API with PingFederate-related policies applied from `PingForDev` to `PingForProd` because both IDPs are of the type `PingFederate`.

franciscoSoler (commented):
nice work, just two small suggestions

chris-campbell9:
Thanks for the review, @franciscoSoler!

priyaayyar (approved):

PR #322, opened 2020-12-28
DOCS-9512 Specify environments available in API Manager

chris-campbell9:
JIRA:
https://www.mulesoft.org/jira/browse/DOCS-9512


modules/ROOT/pages/environments-concept.adoc
```diff
@@ -20,7 +20,7 @@ When you navigate to API Manager, you see the environment control bearing the na
 
 image::environment.png[]
 
-When you click the environment control, a list of your environments appears.
+When you click the environment control, a list of your production and sandbox environments appears.
```
priyaayyar, line 23: You can have many different types of environments created. Are these the only two that will appear?

priyaayyar, line 23:
```suggestion
When you click the environment control, a list of your production and sandbox environments appear.
```

priyaayyar (approved):

modules/ROOT/pages/environments-concept.adoc
chris-campbell9: @priyaayyar 
According to the [Jira](https://www.mulesoft.org/jira/browse/DOCS-9512):
"We had a customer having some issues managing an API since they did not had clear that API Manager only shows Sandbox and Production Environments."

I took this to mean that they are the only two that will appear. Please advise if this is not the case.

chris-campbell9: @priyaayyar 
Isn’t the subject, “a list”, singular, therefore making the verb also singular (“appears”)? I’m thinking the verb isn’t “appear”, as this doesn’t agree with the subject, “a list”.

chris-campbell9: Confirmed: design environments are not displayed in the APIM environment list. I tested this by viewing the environment list in Access Management, noting the various design environments. These same design environments are not present in APIM.

sbehrends: Correct. "Design" environment is not shown in API Manager.

https://docs.mulesoft.com/access-management/environments
> Design environment: Enables you to test and run applications at design time. 

Since API Manager is for operation and not for design time, we don't show such env.

PR #357, opened 2021-04-22
Docs 10518 revamp http caching policy pa

priyaayyar:


modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 | No return codes exist for this policy
 |===
 
-This policy provides a way to store HTTP responses from an API implementation or an API proxy for later reuse. This policy avoids performing multiple calls to the backend when the response of a service does not change often and to optimize against computationally expensive processing. This policy uses the concept of a cache which stores data so future requests for that data can be served faster.
+The HTTP Caching policy enables you to cache HTTP responses for reuse. Caching these responses speeds up the response time for user requests and reduces the load on the backend. For example, if your backend exposes an endpoint for which  the responses to requests are not likely to change, you can reuse the HTTP responses and bypass the backend request processing by using the HTTP Caching policy.
 
-*Note:* Values persisted in Object Store v2 appear as binary, not in plain text.
 
-== What this Policy Stores
+== How this Policy Works
```
kbostrom:
```suggestion
== How This Policy Works
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This policy stores a whole HTTP response from the backend as received by the policy including the status codes, headers, payload, and session variables. If there are policies that edit the response before it is received by this policy, those changes are also present in the response received and therefore stored. If there are policies that make modifications to the response after it passes through the caching policy, those changes won’t be stored, but are replicated each time the value is retrieved from the cache.
+When an HTTP request hits an endpoint, the response flow differs based on the following two basic scenarios:
```
kbostrom: "hits" > "reaches"?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+ 
+* Cache Miss
 
-Each entry has a size limit of 1 MB. In case of HTTP Responses that do not provide content length information, the policy tries to read up to that limit and stores it when it is not reached or continues without storing it, when that limit is exceeded.
+A request that is not present in the cache.
```
kbostrom:
```suggestion
A request that is not present in the cache
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+* Cache Hit 
 
-== How this Policy Works
+A request that is found within the cache 
+
```
kbostrom, line 22: However, I suggest the following:

When an HTTP request reaches an endpoint, what happens next depends on whether that request does not already exist in cache (a _cache miss_) or whether it does (a _cache hit_). 

Is there a reason to describe the miss first?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+=== Cache Miss
+
+When a request is made to an endpoint, the HTTP Caching policy first checks whether the response is already cached. The request reached the backend and the process is initiated only when it's a cache miss. The diagram illustrates the flow of events that occur in this scenario:
```
kbostrom:
```suggestion
In a cache miss scenario, the HTTP Caching policy checks whether a response to the submitted request is already cached:
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
->s|Summary      | Stores HTTP responses from an API implementation
+>s|Summary      | Caches HTTP responses from an API implementation
 >s|Category | Quality of Service
 >s|First Mule version available | v4.1.0
 .1+>.^s| Returned Status Codes
 | No return codes exist for this policy
 |===
 
-This policy provides a way to store HTTP responses from an API implementation or an API proxy for later reuse. This policy avoids performing multiple calls to the backend when the response of a service does not change often and to optimize against computationally expensive processing. This policy uses the concept of a cache which stores data so future requests for that data can be served faster.
+The HTTP Caching policy enables you to cache HTTP responses for reuse. Caching these responses speeds up the response time for user requests and reduces the load on the backend. For example, if your backend exposes an endpoint for which  the responses to requests are not likely to change, you can reuse the HTTP responses and bypass the backend request processing by using the HTTP Caching policy.
```
kbostrom: Remove extra space:
```suggestion
The HTTP Caching policy enables you to cache HTTP responses for reuse. Caching these responses speeds up the response time for user requests and reduces the load on the backend. For example, if your backend exposes an endpoint for which the responses to requests are not likely to change, you can reuse the HTTP responses and bypass the backend request processing by using the HTTP Caching policy.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
-== How this Policy Works
+A request that is found within the cache 
+
+=== Cache Miss
+
+When a request is made to an endpoint, the HTTP Caching policy first checks whether the response is already cached. The request reached the backend and the process is initiated only when it's a cache miss. The diagram illustrates the flow of events that occur in this scenario:
+
+image:http-policy-cache-miss.png[]
+
+. The client sends a request to the API.
```
kbostrom: Note that this step refers to the actors as "the client" and "the API" rather than user request, HTTP request, endpoint, or backend. Using the same terms for the same things throughout a narrative eases comprehension.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The request crosses other policies before HTTP Caching Policy.
```
kbostrom: "crosses" = ?
Does the request skip the other policies or does it undergo processing by those other policies?
```suggestion
. The request proceeds through other policies in the policy chain before reaching the HTTP Caching policy.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The HTTP Caching policy verifies that the request is already in the Object Store. In this case since the request is not present, the Cache Miss process is initiated.
```
kbostrom:
```suggestion
. The HTTP Caching policy checks if the request is already in the object store and, if not, the cache miss process is initiated.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The request continues throughout the chain of other remaining policies.
```
kbostrom, line 32:
```suggestion
. The request continues through the remaining policies in the policy chain.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The request hits the backend and completes all required processing.
```
kbostrom:
```suggestion
. The request reaches the backend and completes all required processing.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The backend returns the response and re-enters the  policy chain.
```
kbostrom:
```suggestion
. The backend responds and the response travels back through the policy chain.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The request reaches the HTTP Caching policy.
+. Caching hits the Object Store again and attempts to save the key of the request for later reuse.
```
kbostrom: "hits the object store again"?
```suggestion
. Caching hits the object store again and attempts to save the key of the request for reuse.
```
This sentence states that "caching hits and attempts to save": do you intend "caching" as the sentence subject?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+
+image:http-policy-cache-miss.png[]
+
+. The client sends a request to the API.
+. The request crosses other policies before HTTP Caching Policy.
+. The HTTP Caching policy verifies that the request is already in the Object Store. In this case since the request is not present, the Cache Miss process is initiated.
+. The request continues throughout the chain of other remaining policies.
+. The request hits the backend and completes all required processing.
+. The backend returns the response and re-enters the  policy chain.
+. The request reaches the HTTP Caching policy.
```
kbostrom: Isn't the "request" now the "response"?

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: request or response? clarify throughout

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. Caching hits the Object Store again and attempts to save the key of the request for later reuse.
+. The request continues throughout the remaining policy chain until it reaches the last one.
```
kbostrom: Didn't the request travel to the backend and then travel back in the opposite direction as a response?
So the request goes through policies A to D, for example, and the response through policies D to A, "returning to the first" in the chain rather than reaching the last?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The response is returned to the client.
 
-When the policy receives a request, it calculates a key under which to store the response. When a second request with the same key arrives, the policy searches the cache for the previously stored response and returns it. If a matching response is found in the cache, the request does not reach either the policies that are defined after the cached one or the API implementation or proxy.
 
-== Where Entries are Stored
+=== Cache Hit
 
-The underlying store used as cache is Mule’s ObjectStore. Depending on the policy’s configuration and where it is deployed, the policy uses an in-memory, persistent, file-based object store or the Runtime Manager ObjectStore v2. In case of the latter, a separate store is created just for the policy with the configured TTL. For more information on accessing the
-object store, see xref:caching-in-a-custom-policy-mule-4.adoc[Caching in a Custom Policy for Mule 4].
+When a request is made to an endpoint, the HTTP Caching policy finds that the response is already cached (cache hit). In this scenario, the request never reaches the backend, and the response to a previously processed request is reused:
```
kbostrom:
```suggestion
In a cache hit scenario, the HTTP Caching policy finds that a response to the submitted request is already cached. The request does not proceed any further in the policy chain toward the backend, and the cached response is reused:
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-== Policy Configuration Parameters
+image:http-policy-cache-hit.png[]
 
-Policy configuration topics:
+. The client sends a request to the API.
+. The request crosses other policies before HTTP Caching policy.
```
kbostrom: ditto edits Line 37

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The HTTP Caching policy verifies that the request is already in the cache.
```
kbostrom, line 51: request or response?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. The key is found in the Object Store and the stored entry is returned.
```
kbostrom:
```suggestion
. The key is found in the object store and the stored entry is returned.
```
stored entry = request or response?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+. Because the key is found in the Object Store and the Cache Hit condition is met, the request does not proceed to the backend. 
+Subsequently, all the policies that must be executed in the policy chain after caching are not executed, in case of a cache hit.
```
kbostrom: This is not a step.
If you use the suggested intro text, you can omit this altogether.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-Policy configuration topics:
+. The client sends a request to the API.
+. The request crosses other policies before HTTP Caching policy.
+. The HTTP Caching policy verifies that the request is already in the cache.
+. The key is found in the Object Store and the stored entry is returned.
+. Because the key is found in the Object Store and the Cache Hit condition is met, the request does not proceed to the backend. 
+Subsequently, all the policies that must be executed in the policy chain after caching are not executed, in case of a cache hit. 
+. The request reverts back through the rest of the policy chain until it reaches the last one. 
+. The response is returned to the client.
```
kbostrom:
```suggestion
. The cached response is returned to the client.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
-* <<Distributed>>
-* <<Persist Cache>>
-* <<Follow HTTP Caching Directives>>
-* <<Invalidation Header>>
-* <<Conditional Request Caching Expression>>
-* <<Conditional Response Caching Expression>>
+
+=== Stored Entries
+
+_Stored entries_ refer to the HTTP responses that are stored in cache. Each entry that is stored has a size limit of 1 MB. The cache can store any Java serializable or input streams. Input streams are read up to the maximum size of 1 MB. If that value is surpassed, then the reading stops and the value is not stored in such a case.
```
kbostrom: So..._stored entries_ = cached responses. Why not introduce this term above?
You could even describe it follow the step in which it is introduced.

kbostrom: copyedit:
```suggestion
_Stored entries_ are the cached HTTP responses. The cache can store any Java serializable or input streams that do not exceed 1 MB; otherwise, the value is not stored in cache.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
 === HTTP Caching Key
 
-The cache works as a dictionary where each response stored in the cache is associated to a string called key. The policy must have a way to compute this key, so this field must contain a DataWeave expression that returns a String value.
+The Cache works as a dictionary, where each response stored in the cache is associated with a string called _key_. For example, the expression #[attributes.headers['key']] uses a header called "key" as the entry key.
```
kbostrom:
```suggestion
The cache works as a dictionary, where each response stored in the cache is associated with a string called _key_. For example, the expression #[attributes.headers['key']] uses the header `key` as the entry key.
```
Again, I think you can work this into the previous texts, which already mention "key" without explaining what that is.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-For example: `#[attributes.headers['key']]`
+=== Cache Size and Entry Expiration
 
-For information on DataWeave variables, see xref:4.1@mule-runtime::dataweave-variables-context.adoc[DataWeave Variables for Mule Runtime] and xref:4.1@mule-runtime::about-mule-message.adoc[Mule Message Structure].
+The cache can hold a specified number of entries at any given time. You can configure this number using the `Maximum Cache Entries` property. Each stored entry is held in the cache memory for a specific period of time, after which the entry expires and must be processed again. This expiration time is called _Time To Live_.
```
kbostrom:
```suggestion
You configure the number of entries that the cache can hold by using the `Maximum Cache Entries` property. Each stored entry is held in the cache memory for a specific period of time, after which the entry expires and must be processed again. This expiration time is called _time to live_.
```
Is an expired entry always reprocessed? Don't you want them to expire and be gone sometimes?

kbostrom: I'm confused: the para above begins with max number of entries and ends with time to live; the following paragraph begins with "conversely" but ALSO begins with max entries. Is this a mistake?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-If the evaluation of the expression results in an error due to missing parameters or any other cause, this policy is not applied and passes the request to the next receptor in the policy chain.
+Conversely, another condition that can trigger the expiration of an entry is when the cache reaches the maximum number of items that it can store. When this scenario occurs, the entry is removed by using the FIFO (First Input First Output) criteria. This means that the earliest entry to reach the cache is removed, even though it hasn't reached the Time To Live value yet.
```
kbostrom: FIFO is “first in first out”

kbostrom: Time To Live > time to live (TTL) as a noun; time-to-live as an adj.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-*Note:*  This parameter is mandatory and it’s default value is `#[attributes.requestPath]`, which means that the key for the caching is the request’s path. If you call the endpoint as `+http://myAppUrl.com/my/policy+` the key is `/my/policy`.
+=== Distributed Cache
 
-=== Maximum Cache Entries
+Each entry in the cache can be shared between different nodes in a cluster or between several workers in Runtime Manager using the `Distributed` option (see table). If this option is not enabled, then each node is assigned its own  cache memory.
```
kbostrom: remove space
```suggestion
Each entry in the cache can be shared between different nodes in a cluster or between several workers in Runtime Manager using the `Distributed` option (see table). If this option is not enabled, then each node is assigned its own cache memory.
```

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: Introducing policy chair here prepares the audience for the following uses of the metaphor.

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: Should we clarify that "completes all required processing" = creating a response?

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: travels back clarifies that the response goes back the way it came rather than restarting from the beginning

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: This is where you could explain what the key is, rather than using the small subtopic below.

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: This is where you could explain that the cached response is known as a stored entry and talk about that here instead of in the small subtopic below.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+
+_Stored entries_ refer to the HTTP responses that are stored in cache. Each entry that is stored has a size limit of 1 MB. The cache can store any Java serializable or input streams. Input streams are read up to the maximum size of 1 MB. If that value is surpassed, then the reading stops and the value is not stored in such a case.
 
 === HTTP Caching Key
 
-The cache works as a dictionary where each response stored in the cache is associated to a string called key. The policy must have a way to compute this key, so this field must contain a DataWeave expression that returns a String value.
+The Cache works as a dictionary, where each response stored in the cache is associated with a string called _key_. For example, the expression #[attributes.headers['key']] uses a header called "key" as the entry key.
 
-For example: `#[attributes.headers['key']]`
+=== Cache Size and Entry Expiration
```
kbostrom, line 66: "entry expiration" = TTL, which I believe your audience already understands?
None of what follows is wrong, but I think that we really needn't explain the max. entries property or the concept of time to live.
OK, no more CU edits; CE only from here.

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: conversely = in a contrasting or opposite way 
I don't think that max entries and TTL qualify as opposites.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+=== Distributed Cache
 
-=== Maximum Cache Entries
+Each entry in the cache can be shared between different nodes in a cluster or between several workers in Runtime Manager using the `Distributed` option (see table). If this option is not enabled, then each node is assigned its own  cache memory.
 
-This is the maximum number of responses that can be present in the cache at the same time. The criteria for removing the responses from the cache when this maximum is surpassed is FIFO. This configuration is ignored in Runtime Manager.
+=== Persistent Cache
 
-This field is also mandatory and has a default value of 10000 entries.
+A persistent cache enables the stored entries in the cache to persist even after the Mule runtime engine (Mule) is restarted.
```
kbostrom: "after" or "if"?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-*Note:* When using the Runtime Manager ObjectStore v2, this is not enforced.
+When you upgrade a Mule version that has the HTTP Caching policy configured to use the persistent store, the policy tries to maintain the entries stored by the previous Mule version. However in a worst case scenario, the entries in the cache are invalidated and the cache is re-populated as and when new requests arrive. This manipulation of the entries in the cache occurs automatically and is shielded from the user.
```
kbostrom: "When you upgrade a Mule version" or "When you upgrade to a Mule version"?

kbostrom: comma and hyphen: "However, in a worst-case scenario, "

kbostrom: I don't think you need both "as" and "when".

kbostrom: The manipulation is not shielded from the user (protecting the manipulation from the user); it is invisible to the user, or the user is shielded from it.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+To obtain more control over the cache, the HTTP Caching policy interprets some of the HTTP directives from the https://tools.ietf.org/html/rfc7234[RFC-7234] protocol by taking the following headers into account:
 
-*Note:* This is a mandatory parameter and its default value is ten minutes (600 seconds).
+* Cache-Control
+* Expires
+* Date
+* Age
 
-=== Distributed
+==== Cache-Control
```
kbostrom: At some point, evaluate if there is a way to reorganize this info to avoid going to four levels of nesting.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-When running a cluster or several workers in Runtime Manager, the values stored in the cache can be shared between the different nodes. To accomplish this, this option must be checked, otherwise each node has its own cache. In the case of Runtime Manager, the object store v2 must be enabled for this option to work.
+The Cache-Control header can exist either in the request or in the response. Possible  values for the header, which can be combined separated by commas, include:
```
kbostrom: If this is syntax, it should be formatted in monospace.

kbostrom: Extra space:
```suggestion
The Cache-Control header can exist either in the request or in the response. Possible values for the header, which can be combined separated by commas, include:
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-If this box is checked the values stored in the cache persist after a runtime restart. If this option is used in Runtime Manager, enable the object store V2 on the application’s configuration in Runtime Manager.
+** no-cache
 
-By default this option is unchecked.
+The response isn’t searched for, but is directly stored in the cache.
+** no-store
 
-*Note:* When you upgrade a Mule version that has the HTTP Caching policy configured to use the persistent store, the policy tries to maintain the entries stored by the previous Mule version. However in a worst case scenario, the entries in the cache are invalidated and the cache is populated again as new requests arrive. This manipulation of the entries in the cache occurs automatically and is shielded from the user.
+The response isn’t stored in the Cache. However, if the response is already present in the Cache, the policy returns the response.
```
kbostrom:
```suggestion
The response isn’t stored in the cache. However, if the response is already present in the cache, the policy returns the response.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-=== Follow HTTP Caching Directives
+* In responses:
 
-To obtain more control over the cache, this policy implemented some of the HTTP directives from the https://tools.ietf.org/html/rfc7234[RFC-7234]. If this box is checked, the policy takes into consideration the following headers to decide which responses to store, when to use the cache, and when  the entry expires.
+* `no-store`, `no-cache`, `private`
 
-By default this box is checked.
+All these values share the same behavior of the response not being stored in the Cache.
```
kbostrom: parallelism with "In requests:"
```suggestion
For each value, the response is not stored in the cache.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-To accomplish this, use the following headers:
+* `max-age=<integer>`, `s-maxage=<integer>`
 
-* <<Cache-Control>>
-* <<Expires>>
-* <<Date>>
-* <<Age>>
+The <integer> value must be replaced with a valid integer, which indicates the time (in  seconds) for which the response can remain in the cache. If both parameters are defined, `s-maxage` takes precedence over `max-age`.
```
kbostrom:
```suggestion
The `<integer>` placeholder must be replaced with a valid integer that indicates the time (in caseconds) for which the response can remain in the cache. If both parameters are defined, `s-maxage` takes precedence over `max-age`. 
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+* `max-age=<integer>`, `s-maxage=<integer>`
 
-* <<Cache-Control>>
-* <<Expires>>
-* <<Date>>
-* <<Age>>
+The <integer> value must be replaced with a valid integer, which indicates the time (in  seconds) for which the response can remain in the cache. If both parameters are defined, `s-maxage` takes precedence over `max-age`. 
 
-==== Cache-Control
+In addition to these headers, If a global Time-to-Live (TTL) value is also defined for the policy, these header values override the global TTL configured in the policy if the time specified in the headers is less than the global TTL time
```
kbostrom:
```suggestion
If a global time-to-live (TTL) value is also defined for the policy, these header values override the global TTL configured in the policy if the time specified in the headers is less than the global TTL.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This header can be present in the request or the response. The values can be combined separated by commas. The values are:
+==== Expires
 
-In the request:
+If present, this header specifies the date by which the entry in the cache expires. If you have specified the `max-age` directive or `s-maxage` directive, this header is ignored. You must define the header value as stated in https://tools.ietf.org/html/rfc1123[RFC-1123].
```
kbostrom: present tense
```suggestion
If present, this header specifies the date by which the entry in the cache expires. If you specify the `max-age` directive or the `s-maxage` directive, this header is ignored. You must define the header value as stated in https://tools.ietf.org/html/rfc1123[RFC-1123].
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This parameter represents how many seconds a response is stored in the cache, and once this time elapses, that the request is removed from the cache. When a new request arrives, the response is not present in the cache, and a new response must be computed.
-When using the policy with Object Store v2 in Runtime Manager, the edition of this value won't be correctly reflected, in this case it's recommended to remove the policy and object store and apply a new one.
+To obtain more control over the cache, the HTTP Caching policy interprets some of the HTTP directives from the https://tools.ietf.org/html/rfc7234[RFC-7234] protocol by taking the following headers into account:
 
-*Note:* This is a mandatory parameter and its default value is ten minutes (600 seconds).
+* Cache-Control
+* Expires
+* Date
+* Age
```
kbostrom: GLOBALLY search for these and capitalize them appropriately: if they are syntax, format them in monospace and exactly mimic the onscreen capitalization. 
In the following texts, these are sometimes capitalized and sometimes not.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-** `no-store,  no-cache, private` - All these values share the same behavior, the response isn't stored in the cache.
-** `max-age=<integer>, s-maxage=<integer>`  - The <integer> value should be replaced with an integer. Indicates how many seconds the response should live in the cache (if both are defined `s-maxage` takes precedence over `max-age`). This value overrides the global TTL configured in the policy.
+==== Age
 
-Example of header present in the request: `Cache-Control: no-cache,no-store`
+The `age` header indicates the time (in seconds) elapsed since the origin of the cached response specified in the date header. This header is calculated by the policy and added to each response that is retrieved from the cache.
 
-This results in the request not being searched with the value from the cache and not storing the result.
+The expiration time is calculated using the cache-control, date, and expires headers. However, if the resulting expiration time exceeds the one imposed by the TTL value of the entry, the cache entry expires anyway.
```
kbostrom: Note that header names are lowercase and not mono. See comment at Lines 93-6.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-Example of header present in the response: `Cache-Control: max-age=2, s-maxage=10`
+=== Invalidate
 
-This results in the response are cached for 10 seconds.
+The `invalidate` header, if configured in the HTTP Caching policy, invalidates the entries in the Cache, thereby causing the request to be processed again. You specify the name of the header in the Invalidation Header configuration to turn on the option. The value of the header can take only one of the following  options:
```
kbostrom: GLOBALLY lowercase "cache" except when quoting syntax or using as part of a proper noun.
```suggestion
The `invalidate` header, if configured in the HTTP Caching policy, invalidates the entries in the cache, thereby causing the request to be processed again. You specify the name of the header in the Invalidation Header configuration to turn on the option. The value of the header can take only one of the following  options:
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+`invalidate`
 
-==== Expires
+This option invalidates the entry whose key matches the current request.
+`invalidate-all`
+This option invalidates all the entries from the Cache.
```
kbostrom:
```suggestion
This option invalidates all the entries from the cache. 
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+The `invalidate` header, if configured in the HTTP Caching policy, invalidates the entries in the Cache, thereby causing the request to be processed again. You specify the name of the header in the Invalidation Header configuration to turn on the option. The value of the header can take only one of the following  options:
+`invalidate`
 
-==== Expires
+This option invalidates the entry whose key matches the current request.
+`invalidate-all`
+This option invalidates all the entries from the Cache. 
 
-This header value should be defined as stated in https://tools.ietf.org/html/rfc1123[RFC-1123].
+For example, if the following values are configured in the policy for the request:
```
kbostrom: This is the beginning of a sentence broken around an example (two, actually). It would be better to reword so that this is not a fragment, something like...:
```suggestion
For example, imagine that the following values are configured in the policy for the request:
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+* `Invalidation header`: "myInvalidationHeader"
 
-==== Date
+The following command invalidates the entry with key “/my/policy" from the cache:
 
-If this header is defined as stated in RFC-1123, it is considered as the time creation of the response. If not defined, the date header is added with the time of reception of the response. This header is used in conjunction with the values defined in the `max-age` and `s-maxage` directives of the Cache-Control header.
+`curl http://myAppUrl.com/my/policy -H “myInvalidationHeader:invalidate”`
 
-==== Age
+Conversely, the following request invalidates all entries from the cache:
```
kbostrom:
```suggestion
The following request invalidates all entries from the cache:
```
Improper use of "converse"

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This header is calculated by the policy and added to each response returned that is retrieved from the cache, it indicates the seconds since the origin of the cached response specified in the date header.
+`curl http://myAppUrl.com/my/policy -H “myInvalidationHeader:invalidate-all”`
 
-The expiration time is calculated using the Cache-Control, Date, and Expiration headers. However, if the resulting expiration time surpasses the one imposed by the <<Entry Time To Live>>, it expires due to this one.
+=== Conditional Caching
 
-=== Invalidation Header
+Conditional caching allows you to configure a set of conditions that must be met before entries are stored in the cache. If none of the conditional caching parameters are configured, the cache stores the responses for every incoming request. If configured, both parameters are evaluated for each request and determined whether the response of the current request must be stored.
```
kbostrom: GLOBAL, per style ref: "allows" > "enables"

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-The expiration time is calculated using the Cache-Control, Date, and Expiration headers. However, if the resulting expiration time surpasses the one imposed by the <<Entry Time To Live>>, it expires due to this one.
+=== Conditional Caching
 
-=== Invalidation Header
+Conditional caching allows you to configure a set of conditions that must be met before entries are stored in the cache. If none of the conditional caching parameters are configured, the cache stores the responses for every incoming request. If configured, both parameters are evaluated for each request and determined whether the response of the current request must be stored. 
 
-This parameter, if defined, indicates the name of the header that will be used for invalidating values in the cache. If not defined, it won’t be possible to invalidate the entries from the cache. The header can take two values:
+In case of the conditional request expression, only the responses for incoming requests with HTTP methods `GET` or `HEAD` are cached by default. In case of the conditional response expression, only status codes specified by https://tools.ietf.org/html/rfc7231#section-6.1[RFC-7231] are cached by default.
```
kbostrom:
```suggestion
In the case of the conditional request expression, only the responses for incoming requests with HTTP methods `GET` or `HEAD` are cached by default. In the case of the conditional response expression, only status codes specified by https://tools.ietf.org/html/rfc7231#section-6.1[RFC-7231] are cached by default.
```

kbostrom: You could replace "In the case of" with "For" in each sentence.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-* `invalidate`: This option invalidates from the cache the entry with the key that the current request has.
-* `invalidate-all`: This option invalidates from the cache all the entries from the cache.
+For more information about the default values, see the Configuring Policy Parameters section.
```
kbostrom:
```suggestion
For more information about the default values, see Configuring Policy Parameters.
```

kbostrom: ...but do we need to xref the very next text block?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-If the header is present in a request, due to the invalidation of at least the key present in the request, the cache won’t be inquired for a previously existing result.
+== Configuring Policy Parameters
```
kbostrom, line 163:
```suggestion
== Table of Policy Parameters
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This parameter is optional and by default it is not defined.
+When you apply the HTTP Caching policy to your API from the UI, you can configure the following parameters:
 
-Example:
+[%header%autowidth.spread,cols="40%,45%, 15%"]
+|===
+| *Parameter* | *Description* | *Required?*
```
kbostrom, line 169: I don't typically see these formatted using * *...?

kbostrom: GLOBALLY lowercase "the cache" in the table; I've not marked each instance.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+| HTTP Caching Key | A DataWeave expression
```
kbostrom:
```suggestion
| HTTP Caching Key | A DataWeave expression. 
```

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: GLOBALLY search for "object store" and use lowercase for generic ("the object store") and init. cap. for CloudHub Object Store. I defaulted to the lowercase, but now I see the proper noun CloudHub Object Store.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+Default value: `#[attributes.requestPath]`
+| Yes
+| Maximum Cache Entries | Specifies the maximum number of entries that can be stored in the Cache at any given time.
+Default value: `10000`
+| Yes
+| Entry Time To Live | Specifies the amount of time (in seconds) after which a single entry expires from the Cache.
+Default value: `600`
+| Yes
+| Distributed | Configures the Cache to be distributed among different nodes in a cluster.
+If you are using CloudHub, ensure that Use Object Store v2 is enabled when deploying an application.
```
kbostrom:
```suggestion
If you are using CloudHub, ensure that Object Store v2 is enabled when deploying an application.
```
```suggestion
If you are using CloudHub, ensure that Use Object Store v2 is enabled when deploying an application.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+Default value: `false`
+| Yes
+| Persistent Cache | Configures the cache to persist between different restarts of the Mule instance.
+If you are using CloudHub, ensure that you enable Use Object Store v2 when deploying an application.
```
kbostrom:
```suggestion
If you are using CloudHub, ensure that you enable Object Store v2 when deploying an application.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+Default value: `false`
+| Yes
+| Follow HTTP Caching directives | Enables the usage of Cache-Control header directives.
```
kbostrom:
```suggestion
| Follow HTTP Caching directives | Enables the use of Cache-Control header directives.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+|===
 
-Considering the following values for the policy, #[attributes.requestPath] for the <<HTTP Caching Key>> and myInvalidationHeader for the invalidation header, the request:
+== FAQ
 
-[source,console]
-----
-curl http://myAppUrl.com/my/policy -H“myInvalidationHeader:invalidate”
-----
+*What happens when the evaluation of the Key results in any error?*
```
kbostrom:
```suggestion
*What happens when the evaluation of the key results in any error?*
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This command invalidates the entry with key “/my/policy" from the cache. On the other hand the request:
+If the evaluation of the expression for a particular request results in an error, the policy takes no effect and the request continues to the next receptor in the policy chain.
```
kbostrom:
```suggestion
If the evaluation of the expression for a particular request results in an error, the policy does not take effect and the request continues to the next policy in the policy chain.
```

modules/ROOT/pages/http-caching-policy.adoc
kbostrom: Why "both parameters"? Are there only two? I would expect "a set of conditions" to be more than two, and we have a whole table full below, right?

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
-This command invalidates all entries from the cache.
+No, but you can apply a policy before this one and transform the response to something else.
 
-=== Conditional Request Caching Expression
+*What happens if one parameter is configured to search for the response in the cache and the other parameter is configured for the opposite action?*
 
-This field holds a DataWeave expression that evaluates the response and returns a boolean. +
-If the request matches the expression, the expression returns True and the policy stores the request in the cache.
-If the request takes any other value, the expression returns False and the cache won’t store the request.
+This is the same as configuring the response to be stored in the cache but has a no-store directive in the Cache-Control header.
```
kbostrom:
```suggestion
This is the same as configuring the response to be stored in the cache when there is a no-store directive in the Cache-Control header.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
-This parameter is optional and its default value is:
 
-[source,mel]
-----
-#[attributes.method == 'GET' or attributes.method == 'HEAD']
-----
+*How does running the HTTP Caching policy different in CloudHub compared with Mule runtime engine?*
 
-This means that only the responses for incoming requests with HTTP methods GET or HEAD are cached by default.
+* If the Distributed option is checked, the cache is always persistent.
```
kbostrom:
```suggestion
* If the Distributed option is selected, the cache is always persistent.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
-For information on DataWeave variables, see xref:4.1@mule-runtime::dataweave-variables-context.adoc[DataWeave Variables for Mule Runtime] and xref:4.1@mule-runtime::about-mule-message.adoc[Mule Message Structure].
+In this situation, the response isn’t stored. You must configure both the Cache-Control header and the cache to be able to store and search for the response in the cache.
 
-This parameter is optional and its default value is:
 
-[source,mel]
-----
-#[attributes.method == 'GET' or attributes.method == 'HEAD']
-----
+*How does running the HTTP Caching policy different in CloudHub compared with Mule runtime engine?*
```
kbostrom:
```suggestion
*How does running the HTTP Caching policy in CloudHub differ compared with running it in Mule runtime engine?*
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
 
-This means that only the responses for incoming requests with HTTP methods GET or HEAD are cached by default.
+* If the Distributed option is checked, the cache is always persistent.
+* The maximum cache configuration is ignored in Runtime Manager.
+* Edition of the TTL of the policy is not correctly reflected while using Object Store V2. To change this value, remove the HTTP Caching policy and apply a new one.
```
kbostrom:
```suggestion
* The edition of the TTL of the policy is not correctly reflected while using Object Store V2. To change this value, remove the HTTP Caching policy and apply a new one.
```
```suggestion
* Edition of the TTL of the policy is not correctly reflected while using Object Store V2. To change this value, remove the HTTP Caching policy and apply a new one.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+Additionally, if you have not configured the request expression or response expression, the cache is used for all requests. All the responses are stored in the cache respectively (expression #[true]).
 
-This parameter is optional and its default value is:
+*Can I modify the Invalidation header values?*
 
-[source,mel]
-----
-#[[200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501] contains attributes.statusCode]
-----
+No, you can only modify the header name.
```
kbostrom:
```suggestion
No, you can modify only the header name.
```
```suggestion
No, you can only modify the header name.
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,414 @@ endif::[]
...
+ 
+*If I deploy my application in an on-premise server, will it use Object Store v2?*
 
-This means that only status codes specified by https://tools.ietf.org/html/rfc7231#section-6.1[RFC-7231] are cached by default.
+No, Object Store V2 is not available for on-premises servers. You must use the Mule Object Store. For more information, see xref:object-store::osv2-faq#can-i-use-object-store-v2-for-apps-deployed-to-on-premises-servers[Object Store V2 FAQs].
```
kbostrom:
```suggestion
No, Object Store v2 is not available for on-premises servers. You must use the Mule object store. For more information, see xref:object-store::osv2-faq#can-i-use-object-store-v2-for-apps-deployed-to-on-premises-servers[Object Store v2 FAQs].
```

kbostrom (dismissed):
CE

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done!  Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: No can't change it now..we have this heading in all other policies as well.

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks! Leaving it as is.

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Request! Thanks! Done!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,419 @@ endif::[]
...
-== Where Entries are Stored
+. The client sends a request to the API.
+//Note that this step refers to the actors as "the client" and "the API" rather than user request, HTTP request, endpoint, or backend. Using the same terms for the same things throughout a narrative eases comprehension.
+. The request proceeds through other policies in the policy chain before reaching the HTTP Caching policy.
+. The HTTP Caching policy checks if the request is already in the object store and, if not, the cache miss process is initiated.
+. The request continues throughout the chain of other remaining policies.
+. The request reaches the backend and completes all required processing.
+. The backend responds and the response travels back through the policy chain.
+. The response reaches the HTTP Caching policy. 
+// (@lorena--the request reaches the policy or the response?)
```
LorenaSoledad: Hi @priyaayyar At this stage, the **response**, because it already hit the backend.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,419 @@ endif::[]
...
+. Caching hits the object store again and attempts to save the <<key,key>> of the request for later reuse.
+//@Lorena--Do we mean that the policy hit the object store again?
```
LorenaSoledad: @priyaayyar Yes, exactly, the policy hits the object store.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,419 @@ endif::[]
...
 
-This is the maximum number of responses that can be present in the cache at the same time. The criteria for removing the responses from the cache when this maximum is surpassed is FIFO. This configuration is ignored in Runtime Manager.
+=== Persistent Cache
 
-This field is also mandatory and has a default value of 10000 entries.
+A persistent cache enables the stored entries in the cache to persist if the Mule runtime engine (Mule) is restarted.
 
-*Note:* When using the Runtime Manager ObjectStore v2, this is not enforced.
+When you upgrade a Mule version that has the HTTP Caching policy configured to use the persistent store, the policy tries to maintain the entries stored by the previous Mule version. However in a worst-case scenario, the entries in the cache are invalidated and the cache is re-populated when new requests arrive. This manipulation of the entries in the cache occurs automatically and is invisible to the user.
+//When you upgrade to a Mule version?--Lorena
```
LorenaSoledad: In CloudHub, for example. You can upgrade your runtime instance manually.

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,419 @@ endif::[]
...
+The Cache-Control header can exist either in the request or in the response. Possible values for the header, which can be combined separated by commas, include:
 
-=== Persist Cache
 
-If this box is checked the values stored in the cache persist after a runtime restart. If this option is used in Runtime Manager, enable the object store V2 on the application’s configuration in Runtime Manager.
+* In requests:
 
-By default this option is unchecked.
+** no-cache
+//These are values right?Lorena
```
LorenaSoledad: @priyaayyar Yes, exactly. `no-cache` and `no-stores` are possible values for the request, while the others below are possible values for the response.

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Ok Thanks!

LorenaSoledad (dismissed):

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Ok done!

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,240 @@ endif::[]
...
-----
-#[attributes.method == 'GET' or attributes.method == 'HEAD']
-----
+`curl http://myAppUrl.com/my/policy -H “myInvalidationHeader:invalidate-all”`
 
-This means that only the responses for incoming requests with HTTP methods GET or HEAD are cached by default.
+== Conditional Caching
 
-=== Conditional Response Caching Expression
+Conditional caching enables you to configure a set of conditions that must be met before entries are stored in the cache. If none of the conditional caching parameters are configured, the cache stores the responses for every incoming request. If configured, parameters are evaluated for each request and determined whether the response of the current request must be stored.
```
kbostrom:
```suggestion
Conditional caching enables you to configure a set of conditions that must be met before entries are stored in the cache. If none of the conditional caching parameters are configured, the cache stores the responses for every incoming request. If configured, parameters are evaluated for each request to determine whether the response of the current request must be stored. 
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,240 @@ endif::[]
...
+Default value: `false`
+| Yes
+| Persistent Cache | Configures the cache to persist between different restarts of the Mule instance.
+If you are using CloudHub, ensure that you enable Object Store v2 when deploying an application.
+Default value: `false`
+| Yes
+| Follow HTTP Caching directives | Enables the use of Cache-Control header directives.
+Default value: `true`
+| Yes
+| Invalidation Header | Name of the header used to invalidate a single entry or the entire Cache. | No
```
kbostrom:
```suggestion
| Invalidation Header | Name of the header used to invalidate a single entry or the entire cache. | No
```

modules/ROOT/pages/http-caching-policy.adoc
```diff
@@ -6,241 +6,240 @@ endif::[]
...
 
-=== What happens if I don’t define some of the optional parameters?
 
-* If invalidation header is not set, it won’t be possible to invalidate the cache in a request.
-* If the request expression or response expression are not set, the cache is used for all requests, and stores all the responses in the cache respectively, as if the expression `#[true]` was set.
+*What happens if I don’t define some of the optional parameters?*
 
-=== Can I change the Invalidation Header values?
+If you have not configured the Invalidation header, you cannot invalidate the cache in a request.
+Additionally, if you have not configured the request expression or response expression, the cache is used for all requests. All the responses are stored in the cache respectively (expression #[true]).
```
kbostrom: What does "respectively" mean here? We typically use it to mean "in the same order" or "to both things in the same order." So, the cache is used only for requests, or for all the requests, first, and then all the responses thereafter?

kbostrom (dismissed):
RA + skim

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

modules/ROOT/pages/http-caching-policy.adoc
priyaayyar: Done! Thanks!

kbostrom (approved):
RA

PR #372, opened 2021-05-23
fix the broken nested item

mtcarvalho:
It is actually shown like this in [docs](https://docs.mulesoft.com/api-manager/2.x/api-proxy-landing-page):

image

Fixed this here.


chris-campbell9 (approved):

chris-campbell9:
Thank you, @mtcarvalho. Good catch!

PR #376, opened 2021-06-04
DOCS-11229 adding some edits

IsaacEldridge:
This PR does a few things: some general pre-kim edits for proper naming convention in titles, adds a few links back to relevant docs in DataGraph, and corrects some naming issues in the nav

(It's not live yet, but I added a section in the DataGraph overview that lists key management tasks that are handled in API Manager)


modules/ROOT/pages/datagraph-landing-page.adoc
```diff
@@ -1,16 +1,20 @@
-= DataGraph Administration
+= Anypoint DataGraph Administration
```
priyaayyar, line 1: Should we call this Configure DataGraph or something like that?

priyaayyar (approved):

modules/ROOT/pages/datagraph-landing-page.adoc
IsaacEldridge: Let me run this by Kim. It's not really "configuring"

PR #437, opened 2022-03-28
added a note on "protected" properties

btrembeczki:
I have had a case where the customer was using one of these properties and when displaying the policy in API Manager, the previously set value does not persist on the UI. However, the value is used for the policy itself, but it is still best if customers don't use this one.


glenn-rodgers-sf:

PR #438, opened 2022-04-13
W-10945772: Fix broken alt-text syntax in images

fermujica:


modules/ROOT/pages/analytics-chart.adoc
```diff
@@ -38,4 +38,4 @@ image::analytics-performance.png[Avg Latency 2.74k]
 
 Enable or disable the tracking of policy violations by checking the Include Policy Violations checkbox on an Analytics chart screen. The chart will show traffic rejected, or not rejected, by a policy.
 
-image::policy-violations-dashboard.png[]
\ No newline at end of file
+image::policy-violations-dashboard.png[]
```
valkyrie69, line 41: If this image is described in the paragraph on line 39, I think we need to reference that--something like... "The chart shows traffic that is rejected or not rejected by a policy, as shown in the following image:"

modules/ROOT/pages/deprecate-api-task.adoc
```diff
@@ -9,7 +9,7 @@ While transitioning consumers of your API to an updated version, you can prevent
 . On the API Administration page, click the API version link.
 . On the API version details page, select Deprecate version from the dropdown.
 +
-image::managing-api-versions-b2d89.png[Deprecate version or Undeprecate version,delete version, export version, Request API access terms & conditions]
+image::managing-api-versions-b2d89.png["Deprecate version or Undeprecate version,delete version, export version, Request API access terms & conditions"]
```
valkyrie69, line 12:
```suggestion
image::managing-api-versions-b2d89.png["Deprecate or Undeprecate version, delete version, export version, Request API access terms & conditions."]
```

valkyrie69 (approved):

PR #440, opened 2022-04-16
corrected proxy setting

franjo2016:
the original value was wrong, and had no effect (at least for mule 4.4 runtime)


glenn-rodgers-sf:

PR #459, opened 2022-08-12
W-11026379 add alt text to images bsj

bjoson-mulesoft:


modules/ROOT/pages/analytics-chart.adoc
```diff
@@ -25,23 +25,23 @@ The top apps summary chart lists the number of requests made to the API by each
...
+image::analytics-performance.png["A latency summary chart showing the average latency for the highest point on the chart, which is 8:30 am."]
 
 == Including Policy Violations in Charts
 
 *Note:* Users must have the View Policies permission set to view policies or edit a Policy Violation alert.
 
 Enable or disable the tracking of policy violations by checking the Include Policy Violations checkbox on an Analytics chart screen. The chart shows traffic rejected, or not rejected, by a policy.
 
-image::policy-violations-dashboard.png[]
+image::policy-violations-dashboard.png["An analytics chart screen with the Include Policy Violations checkbox indicated with a callout."]
```
valkyrie69, line 44: I don't think the quotes are required when the alt text isn't comma-separated, but I don't think it necessarily causes issues to have the quotes either.

valkyrie69 (approved):

modules/ROOT/pages/analytics-chart.adoc
bjoson-mulesoft: Thank you for clarifying!

PR #592, opened 2025-01-14
W-17316074 Managed Flex API

glenn-rodgers-sf:


guidorc (commented):
Looking good! I believe we should also update step 4
Screenshot 2025-01-16 at 2 17 41 PM
Since the gateway statuses will no longer be **Connected** and **Disconnected** but **Running** and **Not Running** instead.

modules/ROOT/pages/_partials/api-configuration-tables.adoc
```diff
@@ -11,7 +11,19 @@
 | *Consumer endpoint* | Specifies a proxy application's address for consumers to use for sending requests.| No |
 | *Client provider* | Specifies a client provider for the API. | No | Anypoint Platform acts as the client provider by default. To configure an external client provider, see xref:access-management::managing-api-clients.adoc[Client Providers].
```
nelak: Client provider is not listed under advanced options for `self-managed`
image

JennyHajee (dismissed):

modules/ROOT/pages/_partials/api-configuration-tables.adoc
marinasasso: Resolved. Github won't let me mark it as resolved

modules/ROOT/pages/create-instance-task-flex.adoc
```diff
@@ -16,7 +16,7 @@ include::partial$task-add-api-instance.adoc[tags=intro1;flex-intro;intro2]
 +
 [NOTE]
 --
-If you do not see a Flex Gateway listed, or you see a Flex Gateway but its status is DISCONNECTED,
+If you do not see a Flex Gateway listed, or you see a Flex Gateway but its status is *Not Running*,
```
Crispy-Salesforce:
```suggestion
If you don't see a Flex Gateway listed, or you see a Flex Gateway but its status is *Not Running*,
```

modules/ROOT/pages/create-instance-task-flex.adoc
```diff
@@ -179,4 +189,4 @@ include::partial$task-import-api.adoc[]
 
 == Notes
 
-* Although OpenAPI Specification (OAS) 3.0 is supported, the callback feature is not. To work around this issue, either handle the callback outside of the Mule runtime engine domain or use an OAS 3.0 specification that does not use callbacks.
\ No newline at end of file
+* Although OpenAPI Specification (OAS) 3.0 is supported, the callback feature is not. To work around this issue, either handle the callback outside of the Mule runtime engine domain or use an OAS 3.0 specification that does not use callbacks.
```
Crispy-Salesforce:
```suggestion
* Although OpenAPI Specification (OAS) 3.0 is supported, the callback feature is not. To work around this issue, either handle the callback outside of the Mule runtime engine domain or use an OAS 3.0 specification that doesn't use callbacks.
```

Crispy-Salesforce (dismissed):

Crispy-Salesforce (approved):

